PR #163, opened 2025-04-16
feat: add webhook service configuration to cluster manager

jokuniew:
### Description

Enabled admission webhook that prevents from cluster-template deletion that is in use

Fixes # (issue)

Enabled webhook feature in cluster-manager. This feature was implemented 2 sprints ago but was disabled due to network issues on orchestrator.
Issues were resolved with https://github.com/open-edge-platform/orch-utils/pull/67

### How Has This Been Tested?

added test case in cluster orch smoke tests to check webhook behaviour

### Checklist:


jokuniew:
webhook funcionality is introduced and validated
now i'll move the webhook enablement in values.yaml to source of cluster-manager helm chart

adorney99 (approved):

PR #181, opened 2025-04-17
CO: enable advanced CAPI providers configuration

jdanieck:
### Description

Before this PR [Cluster API Operator helm chart](https://cluster-api-operator.sigs.k8s.io/02_installation/04_helm-chart-installation) was used to deploy both the CAPI operator and the CAPI providers configuration CRs. 

The helm chart exposes only limited configuration which is not sufficient for our current needs (e.g. we need the ability to provide custom image or configure metrics). Because of that we will use official operator helm chart to deploy operator only and our [custom helm chart](https://github.com/open-edge-platform/orch-utils/pull/76) to configure CAPI providers CRs.

No new 3rd party dependencies.

### How Has This Been Tested?

Deployed Orchestrator from _co-capi-providers-configuration_ branch on Coder.

image
image
image

### Checklist:

### DIFF
--- a/argocd/applications/custom/capi-providers-config.tpl
+++ b/argocd/applications/custom/capi-providers-config.tpl
@@ -46,6 +46,6 @@ controlplane:
 #    deployment:
 #      containers:
 #      - name: manager
-#        imageUrl:  docker.io/andybavier/cluster-api-provider-rke2-controlplane:latest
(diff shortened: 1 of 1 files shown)


togashidm (approved):
lgtm

charlesmcchan (approved):

andybavier (approved):

PR #191, opened 2025-04-18
test: always run get cluster agent logs in vip

jokuniew:
### Description

If UI E2E tests fail - the "Wait for cluster agent to be ready" step is skipped
Later - AO/CO tests are launched. We suspect that enic pod was deleted or restarted in the UI E2E tests

none

### How Has This Been Tested?
CI run

### Checklist:


jokuniew:
"always run cluster agent logs" catched the case, where enic has been restarted 
https://github.com/open-edge-platform/edge-manageability-framework/actions/runs/14577202212/job/40886009948#step:26:16

So the CO/AO tests failed

PR #208, opened 2025-04-22
feat: enabled template controller metrics

adorney99:
### Description

enable metrics in template controller

### DIFF
--- a/argocd/applications/configs/cluster-manager.yaml
+++ b/argocd/applications/configs/cluster-manager.yaml
@@ -5,3 +5,13 @@
 clusterManager:
   extraArgs:
     disable-mt: false
+
+templateController:
+  metrics:
+    service:
+      enabled: true
+      port: 8080
+      labels:
+        app: "template-controller-metrics-svc"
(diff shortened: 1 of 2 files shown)


argocd/applications/configs/cluster-manager.yaml
```diff
@@ -5,3 +5,13 @@
...
+
+templateController:
+  metrics:
+    service:
+      enabled: true
+      port: 8080
+      labels:
+        app: "template-controller-metrics-svc"
+    serviceMonitor:
+      enabled: true
```
jokuniew, line 17:
```suggestion
templateController:
  metrics:
    service:
      enabled: true
    serviceMonitor:
      enabled: true
```

do we need to redefine port and labels once again?

adorney99: good point. Removed anything that's already set by default

jokuniew (approved):
optionally cm version bump to 2.1.0-dev

Paulina-Osikoya (approved):

PR #220, opened 2025-04-23
Use random string for S3-prefix

luoyingf:
### Description

Address the issue of [ITEP-27348](https://jira.devtools.intel.com/browse/ITEP-27348).

Fixes # (issue)

The provisioning script will use a random string of 6 letters as the S3-prefix for new clusters if it is not specified. For existing cluster, it will use the existing S3-prefix.

### How Has This Been Tested?

- Manually tested installation and uninstallation manually from scratch and on a existing cluster.
- Tested it in Jenkins. https://cje-fm-owrp-prod01.devtools.intel.com/nex-nswe-nexnswe/job/OIE-test-automation/job/CD/job/3.0/job/dev-orchestrator/352/console

### Checklist:

### DIFF
--- a/design-proposals/app-orch-dp-import.md
+++ b/design-proposals/app-orch-dp-import.md
@@ -1,0 +1,173 @@
+# Design Proposal: Onboarding Deployment Packages from Github Repositories
+
+Author(s): Scott Baker
+
+Last updated: 2025-04-16
+
+## Abstract
+
+This proposal describes a mechanism for importing a deployment package from an opensource
+application repository into the Orchestrator.
+
+## Problem Statement
+
+The convention chosen for opensource applications wishing to distribute their deployment packages
+is that deployment packages will be placed in yaml format into well-known `deployment-package/` directories. This
+convention has already been adopted for AI Suite applications at Intel
+(ex: [PDD](https://github.com/open-edge-platform/edge-ai-suites/tree/main/manufacturing-ai-suite/pallet-defect-detection/deployment-package)),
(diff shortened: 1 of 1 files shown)


ajaythakurintel (changes requested):
Dummy comment - wait for 3.0 release branches to be created before merging this PR.

pod-configs/utils/lib/common.sh
```diff
@@ -768,3 +773,34 @@ check_s3bucket_exist() {
...
+get_s3_prefix() {
+    # Check if there has been any S3 bucket existing in the Terraform state.
+    existing=$(get_existing_s3_prefix)
+    if [[ -n "$existing" ]]; then
+        echo "$existing"
+        return
+    fi
+
+    # Generate a new one
+    tr -dc a-z0-9 </dev/urandom | head -c 6; echo
```
charlesmcchan, line 804: It is better to use a deterministic hash function so if you wipe out the cluster and redeploy it in same account, region...etc. it will still have the same bucket name. This will make troubleshooting much easier compared to a fully random string that changes all the time.

For example, the bucket name can be something like
`{clustername}-hash(account, region, ...)-loki`
This ensure uniqueness bucket name across account, region...etc. (You can add more hash key if account + region is not sufficient to prevent conflict)
I intentionally left the cluster name outside the hash for readability.

charlesmcchan (approved):

ajaythakurintel (approved):

PR #241, opened 2025-04-29
Bug Fix: Error with exit status 1 during Setup development environment

dmytroye:
### Description

Error with exit status 1 during Setup development environment.

```
Downloading yq from https://github.com/mikefarah/yq/releases/download/v4.34.2/yq_linux_amd64 to /home/ubuntu/.asdf/installs/yq/4.34.2/bin/yq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 -::- -::- -::-     0
100 8740k  100 8740k    0     0  7996k      0  0:00:01  0:00:01 -::- 10.1M
version {version 1.24.1} of golang is already installed
version {version 1.6} of jq is already installed
version {version 1.14.0} of mage is already installed
failed to run download callback: exit status 1
Error: exit status 1
```

Fixes [Error with exit status 1 during Setup development environment](https://jira.devtools.intel.com/browse/ITEP-27330)

n/a

### How Has This Been Tested?

Tested on local bare metal machine with fresh installed Ubuntu 22.04 by running `mage asdfPlugins`

### Checklist:


dmytroye:
image

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
 
 // Install ASDF plugins.
 func AsdfPlugins() error {
+	// Install markdown-lint pre-requisites (jq)
+	if _, err := script.Exec("sudo apt-get install jq -y").Stdout(); err != nil {
+		return fmt.Errorf("error running installing markdown-lint pre-requisites (jq): %v", err)
+	}
```
palade, line 126: We are installing jq as part of the asdf plugins install. It is in the .tool-versions. Why do we need to install it again?

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
...
+	// Install markdown-lint pre-requisites (npm)
+	if _, err := script.Exec("sudo apt-get install npm -y").Stdout(); err != nil {
+		return fmt.Errorf("error running installing markdown-lint pre-requisites (npm): %v", err)
+	}
```
palade, line 130: We are installing nodejs as part of asdf plugins command. This is in .tool-versions. Also I think it also installs npm. I'm wondering if there is a dependency between them, could we swap their position in the list? Let install nodejs first and then markdown-cli2 ?

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
...
+	// Install yamllint pre-requisites (python3-venv)
+	if _, err := script.Exec("sudo apt-get install python3-venv -y").Stdout(); err != nil {
+		return fmt.Errorf("error running installing yamllint pre-requisites (python3-venv): %v", err)
+	}
```
palade, line 134: We have python as part of the asdfplugins. It is not in the .tool-versions. Maybe it also installs python3-venv.

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
...
 	// Install remaining tools
 	if _, err := script.File(".tool-versions").Column(1).
 		MatchRegexp(regexp.MustCompile(`^[^\#]`)).ExecForEach("asdf plugin add {{.}}").Stdout(); err != nil {
-		return err
+		return fmt.Errorf("error running 'asdf plugin add {{.}}': %v", err)
```
palade, line 138: Nit. But can we remove {{.}} from the error message. Have only "error running 'asdf plugin add': %v"

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
...
+		return fmt.Errorf("error running 'asdf install': %v", err)
 	}
 	if _, err := script.Exec("asdf current").Stdout(); err != nil {
-		return err
+		return fmt.Errorf("error running 'asdf current': %v", err)
 	}
 	// Set plugins listed in globalAsdf as global
 	for _, name := range globalAsdf {
 		if _, err := script.File(".tool-versions").MatchRegexp(regexp.MustCompile(name)).Column(2).
 			ExecForEach(fmt.Sprintf("asdf set --home %s {{.}}", name)).Stdout(); err != nil {
```
palade, line 149: "asdf set" <--- this command I think has been added since v0.16 of asdf. It is a command that didn't exist before. I think we should have a check at the beginning of this function to check if v0.16 is installed, otherwise do not run this method since it will fail.

palade (commented):
Left a few comments

mage/Magefile.go
```diff
@@ -120,31 +120,37 @@ func updateEdgeName() {
...
+	// Install markdown-lint pre-requisites (jq)
+	//if _, err := script.Exec("sudo apt-get install jq -y").Stdout(); err != nil {
+	//	return fmt.Errorf("error running installing markdown-lint pre-requisites (jq): %w", err)
+	//}
+	// Install markdown-lint pre-requisites (npm)
+	//if _, err := script.Exec("sudo apt-get install npm -y").Stdout(); err != nil {
+	//	return fmt.Errorf("error running installing markdown-lint pre-requisites (npm): %w", err)
+	//}
+	// Install yamllint pre-requisites (python3-venv)
+	if _, err := script.Exec("sudo apt-get install python3-venv -y").Stdout(); err != nil {
```
se-chris-thach, line 132: This won't work if this runs on Mac. Can we make it work on both OSes?

mage/Magefile.go
dmytroye: Because `jq` doesn't work after installation via asdf (not appears in $PATH or any bin folder) - asdf plugin bug?

se-chris-thach (changes requested):
Is it possible we can also create a dedicated test or something that runs in CI to make sure this Mage target doesn't break going forward? This has been a pain point in the past.

mage/Magefile.go
dmytroye: No, it doesn't

mage/Magefile.go
dmytroye: Removed

mage/Magefile.go
palade: yes it does - I thin it is being installed when nodejs is installed 

```
ubuntu@ubuntu:~$ npm
No version is set for command npm
Consider adding one of the following versions in your config file at
nodejs 23.6.1
```

mage/Magefile.go
palade: It should be in the $PATH. It is made available through the asdf shims `/home/ubuntu/.asdf/shims`

These are set during asdf installation

```
. "$HOME/.asdf/asdf.sh"
. "$HOME/.asdf/completions/asdf.bash"

```

Note that the code above is for v0.15.0. Maybe slightly different for  asdf v0.16.0

mage/Magefile.go
dmytroye: From my point of view is not the best idea to check version of CLI tool in Go code, it would complicate the code unnecessary.

mage/Magefile.go
dmytroye: Good point. But we have much more mage targets that uses Linux commands like `sudo apt-get install` etc.

mage/Magefile.go
dmytroye: Got the point, added Python to the .tool-versions file.

mage/Magefile.go
dmytroye: BTW, this line was removed due to python added to the .tool-version file.

mage/Magefile.go
dmytroye: Got it, fixed

mage/Magefile.go
dmytroye: Changing sequence in the .tool-version doesn't change installation sequence and  markdown-cli2 installation start before nodejs installation and requires npm package manager

palade: There is an ongoing discussion about this that the order is not respected https://github.com/asdf-vm/asdf/issues/929

However, I'm wondering if instead of doing `asdf install`   we can run the same command we run for asdf plugin add. Maybe something like this https://github.com/open-edge-platform/edge-manageability-framework/blob/36f6f314e0429a1243f1c6e4ece4c931c944b4e3/mage/Magefile.go#L125


```
err := script.File(".tool-versions").
		// Filter out comment lines
		MatchRegexp(`^[^\#]`).
		// Process each line
		Echo().
		// Execute a command for each line
		EachLine(func(line string, out *script.Pipe) {
			// Split the line into tool and version
			parts := strings.Fields(line)
			if len(parts) >= 2 {
				tool := parts[0]
				version := parts[1]
				cmd := script.Exec(fmt.Sprintf("asdf install %s %s", tool, version))
				if cmd.Error() != nil {
					fmt.Printf("Error installing plugin %s: %s\n", tool, cmd.Error())
				} else {
					fmt.Printf("Output for %s: %s\n", tool, cmd.String())
				}
			}
		}).Error()

```

mage/Magefile.go
dmytroye: It could be done in a separate PR (seems like task)

mage/Magefile.go
dmytroye: https://github.com/open-edge-platform/edge-manageability-framework/pull/241/commits/9b40c4cf8ce9fcda06ab820554748c1111ba7998 Fixed

mage/Magefile.go
```diff
@@ -120,31 +120,29 @@ func updateEdgeName() {
...
+	}
 	// Install remaining tools
 	if _, err := script.File(".tool-versions").Column(1).
 		MatchRegexp(regexp.MustCompile(`^[^\#]`)).ExecForEach("asdf plugin add {{.}}").Stdout(); err != nil {
-		return err
+		return fmt.Errorf("error running 'asdf plugin add': %w", err)
 	}
-	if _, err := script.Exec("asdf install").Stdout(); err != nil {
-		return err
+	if _, err := script.File(".tool-versions").MatchRegexp(regexp.MustCompile(`^[^\#]`)).ExecForEach("bash -c 'line=\"{{.}}\"; tool=$(echo $line | awk \"{print $1}\"); version=$(echo $line | awk \"{print $2}\"); asdf install $tool $version'").Stdout(); err != nil {
```
palade, line 132: If the approach works, then can use the go implementation directly https://pkg.go.dev/github.com/bitfield/script  

There is a code snippet here https://github.com/open-edge-platform/edge-manageability-framework/pull/241#discussion_r2066732884

palade (changes requested):
LGTM! One small change requested

mage/Magefile.go
dmytroye: Please check https://github.com/open-edge-platform/edge-manageability-framework/pull/241/commits/ac42254d9274557d63fa900796825fe1355f2150

palade (approved):

se-chris-thach (approved):

mage/Magefile.go
```diff
@@ -120,31 +120,45 @@ func updateEdgeName() {
...
+		version := parts[1]
+
+		// Run the asdf install command
+		cmd := fmt.Sprintf("asdf install %s %s", tool, version)
+		if _, err := script.Exec(cmd).Stdout(); err != nil {
+			fmt.Printf("error running '%s': %v\n", cmd, err)
+		}
+		return line
+	}).Stdout(); err != nil {
+		return fmt.Errorf("error running 'asdf install': %w", err)
```
palade, line 149: This code I think can be reduced with 

```
	if _, err := script.File(".tool-versions").
		MatchRegexp(regexp.MustCompile(`^[^\#]\S+\s+\S+$`)).ExecForEach("asdf install {{.}}").Stdout(); err != nil {
		return fmt.Errorf("error running 'asdf install': %w", err)
	}
```

se-chris-thach (approved):

PR #244, opened 2025-04-29
Design proposal on converting Standalone ENs to Managed ENs

osinstom:
### Description

This PR adds design proposal on converting Standalone ENs to Managed ENs.

N/A

### How Has This Been Tested?

N/A

### Checklist:


design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+  note over ctl: After OS reboot or manual cloud-init execution
+
+  ctl->>ctl: Activate Edge Node Agents
+```
+
+**Step 0.** A user already provisioned a set of Standalone ENs following the user guides and decides to scale out.
+They don't need to have direct access to the Edge Orchestrator UI/API (could be different personas on-site vs. remote administrator).
+Users don't need to perform any configuration on the Edge Orchestrator beforehand, but we assume that the Edge Orchestrator supports
+the OS version of the Standalone ENs. The remote administrator configures a special user/role for SEN onboarding.
```
teone, line 112: Is there an assumption that the "user/role for SEN" is associated with a single project that will determine in which tenant the SEN is imported?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+1. An on-site user retrieves user credentials for SEN onboarding from the remote administrator.
+2. An on-site user logs into the node and invokes the CLI tool that should already be installed on the EMT image.
+   The user enters into the interactive session with the CLI tool. 
+3. The CLI tool performs initial OS prerequisites checks - for instance, it can check if the UUID and SN are properly set on the OS. 
+4. A user is prompted for inputs. The user will be asked for the orchestrator FQDN, proxy settings (the already set proxy settings should be presented to the user),
+   and SEN onboarding user credentials. The CLI tool should validate that all required parameters are set and the orchestrator should be reachable at this point.
+   Moreover, the CLI tool can ask for additional input from the user such as which Local Accounts, Site or metadata to configure for a given EN (user must select a configuration that
```
teone, line 120: minor: for any additional input that already exist in the Orchestrator I believe would be beneficial to present an interactive choice to the user, for example: 


This is very import if the persona that uses the CLI tool does not have access to the orchestrator to retrieve the `resourceId`s that they needs to input

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+8. The JWT token is provided to the CLI tool and used in subsequent calls to the Onboarding Manager's gRPC endpoint.
+9. At this point, the CLI tool gathered all user inputs (and validated them) and HW/OS info. It invokes the
+    Onboarding Manager's gRPC endpoint to trigger the onboarding process.
+    Note that the CLI tool should retry if any step above fails on the orchestrator side and the Onboarding Manager should be able to
+    handle partial states.
+10. Onboarding Manager validates the JWT role as it currently does for all gRPC APIs.
+11. Onboarding Manager reads the OS info and queries OS profile from Inventory based on the OS version and OS distro.
+    OS version should uniqely identify the OS profile. Note that this is true for EMT OS profiles, but may not be true for mutable OSes.
+    See open issues for more considerations on the support of mutable OSes.
+12. The Onboarding Manager creates a dedicated cloud-init configuration for SEN. The current cloud-init library can be used.
```
teone, line 138: Are the static IPs (feature planned for 3.1 and ) included in this cloud-init? Are they collected via the CLI or auto-discovered from the system?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+- Keeping the solution OS-independent, i.e., the solution should work on any OS that is supported by the Edge Orchestrator.
+- Reducing the user interaction (i.e., logging into the OS, injecting USB sticks, etc.) with Edge Node machines.
+- Enabling the onboarding process at scale, i.e., the design should enable onboarding of multiple machines at once.
+
+The proposed solution is an MVP approach - some of the above goals may not be fully achieved in the first version of the design,
+but the design doesn't preclude achieving them in the future. 
+
+### Assumptions
+
+- Customers will drive the onboarding process from a local developer machine, with desktop, keyboard and mouse.
```
teone, line 43: Is this still true? Reading below it seems like the CLI is executed directly on the SEN

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+    OS version should uniqely identify the OS profile. Note that this is true for EMT OS profiles, but may not be true for mutable OSes.
+    See open issues for more considerations on the support of mutable OSes.
+12. The Onboarding Manager creates a dedicated cloud-init configuration for SEN. The current cloud-init library can be used.
+    Note that the SEN will only require a subset of current cloud-init that is generated for remote provisioning.
+13. Once OS profile is found and matched, all required HW/OS info provided and cloud-init generated, the OM creates Host and Instance resources.
+    The statuses should be set to Onboarded and Provisioned. Host's desired and current state should be set to `ONBOARDED`.
+    The Instance desired state should be set to `RUNNING`, while the current state to `UNSPECIFIED` (until BM agents are up and running).
+14. Once the onboarding is completed, the CLI tool receives the generated cloud-init configuration.
+15. The CLI tool saves the cloud-init under the standard path that is used by EIM.
+    The user can either run the cloud-init manually or reboot the system to trigger the cloud-init.
```
teone, line 145: Can the CLI prompt the user to reboot the system (or invoke cloud-init directly) to complete the process?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+   The user enters into the interactive session with the CLI tool. 
+3. The CLI tool performs initial OS prerequisites checks - for instance, it can check if the UUID and SN are properly set on the OS. 
+4. A user is prompted for inputs. The user will be asked for the orchestrator FQDN, proxy settings (the already set proxy settings should be presented to the user),
+   and SEN onboarding user credentials. The CLI tool should validate that all required parameters are set and the orchestrator should be reachable at this point.
+   Moreover, the CLI tool can ask for additional input from the user such as which Local Accounts, Site or metadata to configure for a given EN (user must select a configuration that
+   already exists on the orchestrator and OM should be responsible for validating if that configuration exists in the Inventory).
+   Also, the CLI tool retrieves hardware info (UUID, Serial Number), OS info (OS version and distro from `/etc/os-release`), current the Secure Boot and Full-Disk Encryption settings,
+   MAC/IP address of the management interface.
+5. The CLI tool downloads the CA certificate from the Edge Orchestrator.
```
rranjan3, line 124: Even to start with, we could move the certificate procurement (to provisioning machine) to an offline mechanism & put into the SEN rather than SEN making an outbound call. Implicit trust has always been questionable.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+  note over ctl: CLI triggered
+
+  ctl->>ctl: Checks local OS prerequisites
+
+  ctl->>ctl: Take user inputs and validate
+
+  ctl->>+pa: Get orchestrator CA certificate
+  pa-->>-ctl: [CA certificate]
+
+  ctl->>+kc: Retrieve JWT token for standalone onboarding user
```
rranjan3, line 88: Why can we not use NIO registration instead of obtaining JWT using credentials here? The NIO flow does trust the deviceinit. Should reduce user interaction. I reckon security is not a concern as we allow NIO with both SB enabled/disabled.
OR
Since the provisioning machine is already well positioned to help user authenticate with the orch,
1.) can pull a bundle from the orch consisting of all artifacts needed + a one-time credential
2.) dump the bundle to the SEN to automate the rest of the flow.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+  end
+
+  note over ctl,user: User installs Standalone EN, decides to scale out
+
+  note over pa,inv: OS profiles created beforehand, no Host/Instance pre-registration
+
+  user->>user: Retrieves standalone onboarding user credentials 
+  user->>ctl: Logs in to the EN and invokes CLI tool
+
```
krishnajs, line 78: We should also add the sequence diagram for the bulk onboard of EMT-S Nodes to the orch.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+  ctl->>+kc: Retrieve JWT token for standalone onboarding user
+  kc->>-ctl: [JWT token]
+
+  ctl->>+om: Invoke OnboardStandaloneService gRPC (UUID, Serial Number, OS info, ...)
+
+  om->>om: Validate JWT standalone onboarding role
+
+  om->>inv: Query Inventory for OS resource by OS version and distribution
+
+  om->>om: Generate cloud-init for Standalone EN
```
krishnajs, line 97: Maybe we need to specific generate cloud init that is needed for the EN agents to work with orch. This ensured the cloud init will not overwrite/undo what user might have configured on the EN.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+  pa-->>-ctl: [CA certificate]
+
+  ctl->>+kc: Retrieve JWT token for standalone onboarding user
+  kc->>-ctl: [JWT token]
+
+  ctl->>+om: Invoke OnboardStandaloneService gRPC (UUID, Serial Number, OS info, ...)
+
+  om->>om: Validate JWT standalone onboarding role
+
+  om->>inv: Query Inventory for OS resource by OS version and distribution
```
krishnajs, line 95: what if the OS version does not match the OS resource ?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+**Step 0.** A user already provisioned a set of Standalone ENs following the user guides and decides to scale out.
+They don't need to have direct access to the Edge Orchestrator UI/API (could be different personas on-site vs. remote administrator).
+Users don't need to perform any configuration on the Edge Orchestrator beforehand, but we assume that the Edge Orchestrator supports
+the OS version of the Standalone ENs. The remote administrator configures a special user/role for SEN onboarding.
+
+1. An on-site user retrieves user credentials for SEN onboarding from the remote administrator.
+2. An on-site user logs into the node and invokes the CLI tool that should already be installed on the EMT image.
+   The user enters into the interactive session with the CLI tool. 
+3. The CLI tool performs initial OS prerequisites checks - for instance, it can check if the UUID and SN are properly set on the OS. 
+4. A user is prompted for inputs. The user will be asked for the orchestrator FQDN, proxy settings (the already set proxy settings should be presented to the user),
```
krishnajs, line 118: I was thinking... we need to provide a UX where customer can download a payload from the Orch instance (like what we download certs today). This payload can have certs, config and other artifacts for EMT-S node to basically connect to Orch (even OTP can be part of it). When customer copies this payload to all ENs (e.g. bulk scp) this agent on the EN (e.g. node agent) can start to communicate with Orch and move to normal operations.

krishnajs (commented):
Similar to BIT + pre-reg will be default for the EIM onboarding and provisioning, we should make Build-EMT-Sn-Import as default. Granularity is 1 EMT-S node... but in general multiple EMT-S Node import.

design-proposals/eim-standalone-managed-en.md
osinstom: Yes, the same assumption we have now for IO onboarding

design-proposals/eim-standalone-managed-en.md
osinstom: yeah, I've thought about that, but we would need to expose API to Inventory on the southbound interface. Or let CLI communicate with northbound API too, which could be feasible but would probably need a different role. Thoughts @daniele-moro @pierventre  ?

design-proposals/eim-standalone-managed-en.md
osinstom: That's a different cloud-init. This is a default cloud-init that EIM uses for EN provisioning. The other per-EN config can be provided via another cloud-init - @niket-intc is working on the design.

We can make CLI collect what day0 configuration should be applied (local accounts, per-EN config, sites, etc.) as user inputs

design-proposals/eim-standalone-managed-en.md
osinstom: Kind of yes - I still assume that users do SSH into ENs and there is no peripherals attached to ENs

design-proposals/eim-standalone-managed-en.md
osinstom: Yes, definitely doable

design-proposals/eim-standalone-managed-en.md
osinstom: That's true, but may require a bit more manual steps, which here the CLI automates that.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Assumptions
+
+- Customers will drive the onboarding process from a local developer machine, with desktop, keyboard and mouse.
+- The local developer machine will have direct access to ENs via local subnet. Customers can SSH into the ENs.
+- SENs being onboarded won't be equipped with desktop, keyboard or mouse. The only machine equipped with peripherals will be
+  the local developer/admin machine.
+- The design will re-use the current APIs of Onboarding Manager to drive IO/NIO-based onboarding.
+  We may just require small modifications to the Onboarding Manager and the gRPC interface.
+
```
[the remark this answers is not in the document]
osinstom: Actually I was trying to make the design OS-independent so that we could support bring your own OS model. We have too much logic that is only used for a single OS distro (like different Tink workflows for EMT and Ubuntu) - I wanted to avoid that.

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+### Solution
```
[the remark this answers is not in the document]
osinstom: Isn't that part of Standalone EN ? I kind of assumed that day0 status is all agents are downloaded and installed, but cannot communicate. 

Do you think we should cover "standalone Ubuntu" case here as well?

design-proposals/eim-standalone-managed-en.md
pierventre: I believe we want just separate from the host/onboarding roles

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+Author(s): Tomasz Osiński
+
+Last updated: 29.04.2025
+
+## Abstract
+
+A Customer Journey for Open Edge Platform assumes that customers can manually deploy a set of Standalone
+Edge Nodes (SEN) that can be onboarded to the Edge Orchestrator at later stage, once a customer is ready to scale their deployment.
+SENs are converted to managed Edge Nodes which, once onboarded, are fully owned by the Edge Orchestrator - customers
+can manage them (e.g., install clusters, applications or perform Day2 OS updates) through the Edge Orchestrator UI and API.
```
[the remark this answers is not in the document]
pierventre: if we build these images - I expect that cluster is able to import those clusters. Is that correct?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+The proposed solution is an MVP approach - some of the above goals may not be fully achieved in the first version of the design,
+but the design doesn't preclude achieving them in the future. 
+
+### Assumptions
+
+- Customers will drive the onboarding process from a local developer machine, with desktop, keyboard and mouse.
+- The local developer machine will have direct access to ENs via local subnet. Customers can SSH into the ENs.
+- SENs being onboarded won't be equipped with desktop, keyboard or mouse. The only machine equipped with peripherals will be
+  the local developer/admin machine.
+- The design will re-use the current APIs of Onboarding Manager to drive IO/NIO-based onboarding.
```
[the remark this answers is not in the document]
pierventre: there is no documentation around IO/NIO. The only thing we document is the pre-registration. IO is always treated as failure scenario - fallback option

design-proposals/eim-standalone-managed-en.md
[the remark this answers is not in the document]
pierventre: or maybe they are just disabled and we cna create later the config and then restart - ubuntu scenario

design-proposals/eim-standalone-managed-en.md
pierventre: yeah we should clrify what is needed for

design-proposals/eim-standalone-managed-en.md
pierventre: we should just return an error to the caller - so in generl we can always retry

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+  ctl->>+om: Invoke OnboardStandaloneService gRPC (UUID, Serial Number, OS info, ...)
+
+  om->>om: Validate JWT standalone onboarding role
+
+  om->>inv: Query Inventory for OS resource by OS version and distribution
+
+  om->>om: Generate cloud-init for Standalone EN
+
+  om->>inv: Create Host and Instance resource, Set EN as Onboarded and Provisioned
```
pierventre, line 99: current_state. I still believe we should force for a top-down reg as pre-requ so we have always desired-current state evolving together

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+**Step 0.** A user already provisioned a set of Standalone ENs following the user guides and decides to scale out.
+They don't need to have direct access to the Edge Orchestrator UI/API (could be different personas on-site vs. remote administrator).
+Users don't need to perform any configuration on the Edge Orchestrator beforehand, but we assume that the Edge Orchestrator supports
+the OS version of the Standalone ENs. The remote administrator configures a special user/role for SEN onboarding.
+
+1. An on-site user retrieves user credentials for SEN onboarding from the remote administrator.
+2. An on-site user logs into the node and invokes the CLI tool that should already be installed on the EMT image.
+   The user enters into the interactive session with the CLI tool. 
+3. The CLI tool performs initial OS prerequisites checks - for instance, it can check if the UUID and SN are properly set on the OS.
```
pierventre, line 117: do we plan to unify around the single CLI tool? Or is this a separate CLI ?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+    handle partial states.
+10. Onboarding Manager validates the JWT role as it currently does for all gRPC APIs.
+11. Onboarding Manager reads the OS info and queries OS profile from Inventory based on the OS version and OS distro.
+    OS version should uniqely identify the OS profile. Note that this is true for EMT OS profiles, but may not be true for mutable OSes.
+    See open issues for more considerations on the support of mutable OSes.
+12. The Onboarding Manager creates a dedicated cloud-init configuration for SEN. The current cloud-init library can be used.
+    Note that the SEN will only require a subset of current cloud-init that is generated for remote provisioning.
+13. Once OS profile is found and matched, all required HW/OS info provided and cloud-init generated, the OM creates Host and Instance resources.
+    The statuses should be set to Onboarded and Provisioned. Host's desired and current state should be set to `ONBOARDED`.
+    The Instance desired state should be set to `RUNNING`, while the current state to `UNSPECIFIED` (until BM agents are up and running).
```
pierventre, line 142: are you sure about this ? I feel we should modify the hrm to touch only the modern status and avoid the modification of the current state. I feel the current state should be own only by the OM

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+Both approaches require having a local daemon scanning for USB sticks or on-disk config files. A local daemon is more
+"heavy" solution as it would require special security permissions and a more complex lifecycle management via systemd.
+It is also less interactive, so that possible intermediate errors may be harder to debug (on the contrary the CLI tool guides users through process).
+
+Both options are still valid for future releases, but for now we decided to go with the CLI tool as the MVP solution.
+
+### Considerations on the user workflow
+
+There are two major workflows we support - bottom-up (IO) and top-down (NIO, requires EN pre-registration).
```
pierventre, line 167: we should not use IO/NIO. Let's substitute with just EMT-S onboarding

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+In this ADR we selected to use the bottom-up approach as it requires less manual steps - user logs in to EN and
+run the entire workflow from the EN itself, without the need to access the Edge Orchestrator UI/API.
+
+Also, the IO flow doesn't require any modifications to UI.
```
pierventre, line 172: can you elaborate here ?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Considerations on scaling the SEN onboarding
+
+The current design assumes that the user will SSH into a Standalone EN and trigger onboarding one by one (it can still be automated by a script).
+Depending on the customers' requirements we can provide a kind of a "Bulk Onboarding Tool" that will allow to onboard multiple ENs at once.
```
pierventre, line 177: or maybe we can automate with ansible?

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+
+### How to map local OS users to Local Accounts?
+
+For now, we won't import existing OS users as local accounts. If we support NIO in the future,
+we can add support for defining local accounts that should be configured on the pre-provisioned SEN that is onboarded.
```
pierventre, line 182: you can create the default-local account though if it is configured for the project

design-proposals/eim-standalone-managed-en.md
```diff
@@ -0,0 +1,236 @@
...
+## Open issues (if applicable)
+
+- The OS version of Standalone ENs deployed by customers may not have a corresponding OS Profile in the Edge Orchestrator.
+  For instance, the customer may have deployed a Standalone EN with custom EMT image that never existed in the Edge Orchestrator.
+  Another example is a customer that has deployed a Standalone EN with an old OS image that is not supported by the Edge Orchestrator 
+  (which has already been upgraded to use newer versions). Yet another issue is that the OS version from `/etc/os-release`
+  will not uniquely identify the OS image if that's a mutable OS (Ubuntu case) and Onboarding Manager is unable to query or validate mutable OS profile
+  based on info provided from SENs.
+  There are possible solutions to this:
+  - Users should be able to create their own OS Profiles that uses a custom OS image that was used for Standalone ENs.
```
pierventre, line 224: shall we put some guardrails in the OS_profile ? For example prevent the provisioning or a/b updates based on the type of the profile

pierventre: maybe add STANDALONE_OS as type and ofrce the user to use that

design-proposals/eim-standalone-managed-en.md
pierventre: I agree - this is something that was discussed long time ago with PDM too when they wanted to opt-in for a lightweight FDO

design-proposals/eim-standalone-managed-en.md
pierventre: I really like this more than IO - that mimics in somehow the FDO flow

design-proposals/eim-standalone-managed-en.md
pierventre: we can use the nb-apis. not sure why we would need a different role though

design-proposals/eim-standalone-managed-en.md
pierventre: do we really want this? CLI is becoming like an orchestrator. Let's focus on the reqs only

PR #246, opened 2025-04-29
ADR for Orchestrator CLI tool

scottmbaker:
### Description

This is a design for an orchestrator-wide CLI.

### Checklist:


design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+- `cli -p acme apply -f nginx.yaml` ... create the objects contained inside nginx.yaml.
+
+- `cli logout` ... log out.
+
+### Miscellaneous Guidelines
+
+- The CLI seeks to minimize the boundaries between teams and components. For example, the following approaches
+  are not desirable:
+
+  - `cli update app-orch application foo` ... app-orch is a team/subsystem designator that is not relevant to
```
adibrastegarnia, line 158: I am wondering if we should consider the potential of adding groups/subgroups like cloud provider CLIs. Let's say we want to perform start/stop VM operations; what does the command look like?"

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+Last updated: 2025-04-29
+
+## Abstract
+
+This ADR describes the design of a command line interface (CLI) for the
+Orchestrator. The CLI is a cross-component tool that may be used from a Linux, Windows, or
+Mac environment, such as from the administrator's laptop. The CLI may also may be leveraged
+as a building block in other environments such as CI/CD or testing.
+
+The CLI described here uses the `Catalog CLI` as a strating point and inspiration. However, this
```
badhrinathpa, line 14: typo: starting

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+#### Verb-Specific and/or Noun-Specific options
+
+Verbs and nouns may have addional non-global options as necessary. For example, the `create application`
+verb/noun pair includes the options `--chart-name` and `--chart-registry` and `--chart-version`.
+
+#### Examples
+
+- `cli login sample-project-admin` ... log in as a user.
+
+- `cli -p acme list applications nginx` ... list all nginx applications in a tabular format.
```
zdw, line 137: Does the CLI tooling support using a go template to format output?

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `cli -p acme list applications nginx =o yaml` ... list all nginx applications and emit the output as yaml.
```
zdw, line 139: `-o` ?

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+  TO-DO: Consider treating completion as a verb/noun operation? i.e. `cli get completion bash`.
+
+- `upload <directory>` ... upload items to the Orchestrator. This is currently a catalog-specific
+  operation and uploads a directory of yaml files to the orchestrator.
+
+- `version` ... return the version number of the CLI.
+
+- `watch` ... watch an object in the Orchestrator using webhook or stream interface, returning changes
+  as they occur.
+  TO-DO: drop this?
```
zdw, line 84: Someone could just wrap with the `watch` command, so would be in favor of dropping

Not all orchestrator interfaces have webhook/stream interfaces, and this could be expensive on the API side.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `wipe` ... wipe all data in a project.
+  TO-DO: drop this? Should be implemented as part of `cli delete project`.
+
+#### Nouns
+
+This ADR does not seek to exhaustively list the set of nouns that are available on the CLI. Each
+component shall add their own nouns as appropriate. For example,
```
zdw, line 92: What does this look like in terms of code structure? One noun per directory/go package?

Are there any nested nouns?  I'd prefer that we avoid this, but in some cases, like "profile" (as we have "App Profiles" and "OS Profiles") this may be unavoidable, or we should use the full name with a prefix.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### Global Options
+
+- `-h` / `--help` ... return help. The help returned may be contextualized to the verb or noun that is
+  being used. For example `cli -h` returns global help whereas `cli create application -h` return help
+  that is relevant to creating applications.
+
+- `-n` / `--noauth` ... disables authentication checks, only useful in a devleopment environment.
+
+- `-p` / `--project` ... sets the project that will be used for the current verb. Required for verbs that
```
Andrea-Campanella, line 121: shall we allow configuration of this ? meaning that somebody can set this once he has logged it. a step further will be automating this via the token/access. if the user has access to only one project no need to go through this step.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Code structure
+
+The CLI shall be implemented in a modular way.
+
+Each noun shall be placed in a file `<noun>.go` in the `internal/cli` directory. For example,
+
+- `internal/cli/application.go` ... implements the noun `application` and its associated verbs.
+
+TO-DO: Here is may make sense to separate code by subsystem for easier maintainability. For example,
```
Andrea-Campanella, line 216: yes, I'd agree to this, although `catalog` might not be my choice, but rather `infra`, `app`, `cluster`, `common`

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+`internal/cli/catalog/application.go`, `internal/cli/cluster/cluster-templates.go`, etc. Discuss.
+
+## Open issues (if applicable)
+
+- Session / Context management. It would be convenient if the CLI could store the context for
+  multiple orchestrator sessions and then easily switch between them.
+
+- Distribution of Binaries. Proposal is to make them a github release, so they may be easily
```
Andrea-Campanella, line 224: to me this is not an open issue but a fact.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+  downloaded by an end-user. Additional scanning or signing of binaries may be necessary.
+
+- Filtering and Naming. This proposal adopts the Catalog convention where most objects are named by
```
Andrea-Campanella, line 227: @damiankopyto @zdw  @pierventre  please provide thoughts here.

Andrea-Campanella (commented):
great start, some food for thought.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+The CLI implements a set of verbs, a set of nouns, and a set of common options that may
+be used to interact with the orchestrator.
+
+The CLI follows a `verb - noun - subject` pattern. For example,
+
+- `cli get application nginx`. The verb is `get`. The noun is `application` and the subject
+  is `nginx`. This particular CLI command would return the application called nginx from the
+  Orchestrator and print them to the console in a human-readable format.
+
+### Syntax
```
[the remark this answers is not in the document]
palade: Maybe `orch` ?

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `login <user-id>` ... log in to the Orchestrator. The CLI shall prompt for a password and retrieve
+  a refresh token. The refresh token is cached locally until it is expired or a `logout` verb is used.
+
+- `logout` ... log out of the Orchestrator. Any active login state, including the refresh token, is
+  discarded.
+
+##### CRUD Operations
+
+- `list` ... lists objects on the Orchestrator. The `list` verb produces a list of all objects
```
palade, line 49: What are objects in this context: components/services deployed on the orchestrator or clusters/applications/ens managed by the orchestrator?

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+The CLI described here uses the `Catalog CLI` as a strating point and inspiration. However, this
+ADR assumes that syntax and semantics of the CLI may diverge from the existing `Catalog CLI` and
+this ADR is not constrained by that existing implementation. As such, examples given in this
+ADR may not necessarily work with the existing catalog CLI.
+
+## Proposal
+
+The CLI implements a set of verbs, a set of nouns, and a set of common options that may
+be used to interact with the orchestrator.
```
palade, line 22: I think we will need to outline what are the differences between this new cli tool and existing mage targets i.e., if new functionality is added is implemented as a mage target or in the cli

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `list` ... lists objects on the Orchestrator. The `list` verb produces a list of all objects
+  on the Orchestrator, with an optional filter. List should return tabular summary-level information
+  about the objects.
+
+- `get` / `describe` ... retrieves objects from the Orchestrator. `get` is for retrieving single objects whereas
+  `list` is for retrieving multiple objects. `get` should retrieve more verbose information about the
+  object than `list`.
+
+- `create` ... creates an object on the Orchestrator. Additional noun-specific parameters are used to
```
damiankopyto, line 57: @scottmbaker  Is this CRUD/verb list exhaustive?
Ie. The `create` action for adding a host is more complex create could consist of
`register` host, `onboard` host, `provision` host - as individual action or in combination or in bulk.

Can/should the list be expandable with component specific verbs ie, `cli register host myhost --somearguments`
or do we need to manage this somehow via noun level options `cli create host myhost --register --onboard --provision --somearguments`

On another hand we may want to create hosts in bulk with autogenerated hostnames (current BIT), is it acceptable to have no (or dummy) subject i.e `cli create host --fromcsv ./my_csv_file.csv`

@zdw @pierventre

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+  describe the properties of the object to be created.
+
+- `delete` ... deletes an object from the Orchestrator.
```
damiankopyto, line 60: Similar question as above deleting a host could cover deauthorize and delete, or force delete, or deauthorization may need to be done on its own without deleting - any thoughts if it should be done on verb or noun level

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `set` / `update` ... modifies one or more values in an existing object on the Orchestrator.
+
+- `apply` ... create or modify an object from a yaml or json specification, from stdin or from a local
+  file.
+
+##### Configuration and utility
+
+- `config` ... configures the CLI. For example, `cli config set endpoint https://my-orchestrator.com/`
```
SeanCondon, line 69: `config` could be handled with `set-config` and `get-config` verbs.

It is about configuring the CLI and not sent to the backend. Settings are stored in the config file as mentioned below.

design-proposals/orch-cli.md
SeanCondon: I think we should also allow listing projects and their details and the org details.

design-proposals/orch-cli.md
SeanCondon: I think the idea is that EIM team should define and own behaviour of EIM related commands as they add their plugin. App Orch will also be a plugin and we will own that. I think we should not go too low on details here, just facilitate the infrastructure and philosophy

design-proposals/orch-cli.md
SeanCondon: I think there should be a modular plugin structure so that EIM can own their part while App Orch can own ours.
Scott has more on code details below

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+- `-h` / `--help` ... return help. The help returned may be contextualized to the verb or noun that is
+  being used. For example `cli -h` returns global help whereas `cli create application -h` return help
+  that is relevant to creating applications.
+
+- `-n` / `--noauth` ... disables authentication checks, only useful in a devleopment environment.
+
+- `-p` / `--project` ... sets the project that will be used for the current verb. Required for verbs that
+  act on projects.
+
+- `-v` / `--verbose` ... enables verbose output.
```
SeanCondon, line 124: I would like to see an ability to print out full details of an API calls made - maybe this can be done through `-vv` as is common with other tools

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+Application Orchestration, Cluster Orchestration, EIM, Platform.
+
+## Implementation plan
+
+The CLI is to be implemented in the `go` programming language, using the popular `cobra` and `viper`
+go libraries.
+
+The CLI shall include a Makefile and github actions that facilitate building it for Linux, Windows, and
+Mac.
```
SeanCondon, line 206: we should thing about package **signing** for these OS's - may need some CI tooling

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+ADR may not necessarily work with the existing catalog CLI.
+
+## Proposal
+
+The CLI implements a set of verbs, a set of nouns, and a set of common options that may
+be used to interact with the orchestrator.
+
+The CLI follows a `verb - noun - subject` pattern. For example,
+
+- `cli get application nginx`. The verb is `get`. The noun is `application` and the subject
```
pierventre, line 26: I am thinking this can collide with argo CRDs - shall we use more specific nouns ?

design-proposals/orch-cli.md
pierventre: so maybe we should spend some words explaining the plugin architecture and clarify the boundaries between the common CLI and the plugins

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+  object than `list`.
+
+- `create` ... creates an object on the Orchestrator. Additional noun-specific parameters are used to
+  describe the properties of the object to be created.
+
+- `delete` ... deletes an object from the Orchestrator.
+
+- `set` / `update` ... modifies one or more values in an existing object on the Orchestrator.
+
+- `apply` ... create or modify an object from a yaml or json specification, from stdin or from a local
```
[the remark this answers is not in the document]
pierventre: we can potenttially use this for our bulk tool. I believe pply is generally understood as way to create an object from stdin

se-chris-thach (commented):
Since we're [unifying the Install/Upgrade process](https://github.com/intel-innersource/documentation.edge.one-edge.maestro/pull/793/) of Orchestrator into a single binary, what if we took it a step further a merged it for the this Orchestrator CLI as proposed in this ADR?

How about a subcommand like `cli install service aws`? This will trigger the install TUI on fresh installs.

@charlesmcchan thoughts?

design-proposals/orch-cli.md
scottmbaker: fixed.

design-proposals/orch-cli.md
scottmbaker: Not at this time -- I think working on an output formatter would be a good idea, and good to get started on it sooner rather than later, before everyone hand-codes their output code. I'm reminded of the one that was in voltctl, but it's been about 5 years. I'm sure there are better things now.

design-proposals/orch-cli.md
scottmbaker: fixed

design-proposals/orch-cli.md
scottmbaker: `watch` was originally designed to exercise/test a websocket-based interface to the catalog that was going to be used to make the GUI event-driven. Let's drop it.

design-proposals/orch-cli.md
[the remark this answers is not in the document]
scottmbaker: Also throwing in `oc` for consideration. Shorter seems to be a trend lately.

design-proposals/orch-cli.md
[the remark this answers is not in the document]
scottmbaker: examples are further down in the ADR in the `Examples` section. Update was intended to update a single object. Not sure if bulk-import should be a separate operation.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+  and we should consider discarding it in favor treating configuration as any other verb-noun operation,
+  i.e. `cli update config --endpoint https://my-orchestrator.com/`.
+
+- `completion` ... return a script that can be used to configure autocompletion.
+  TO-DO: Consider treating completion as a verb/noun operation? i.e. `cli get completion bash`.
+
+- `upload <directory>` ... upload items to the Orchestrator. This is currently a catalog-specific
+  operation and uploads a directory of yaml files to the orchestrator.
+
+- `version` ... return the version number of the CLI.
```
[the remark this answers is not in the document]
scottmbaker: We'll need an endpoint on the orchestrator to return that version number, but it would be useful.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `watch` ... watch an object in the Orchestrator using webhook or stream interface, returning changes
+  as they occur.
+  TO-DO: drop this?
+
+- `wipe` ... wipe all data in a project.
```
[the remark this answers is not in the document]
scottmbaker: dropping.

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+- `config` ... configures the CLI. For example, `cli config set endpoint https://my-orchestrator.com/`
+  TO-DO: This command inherited from the catalog CLI does not follow consistent syntax
+  and we should consider discarding it in favor treating configuration as any other verb-noun operation,
+  i.e. `cli update config --endpoint https://my-orchestrator.com/`.
+
+- `completion` ... return a script that can be used to configure autocompletion.
+  TO-DO: Consider treating completion as a verb/noun operation? i.e. `cli get completion bash`.
+
+- `upload <directory>` ... upload items to the Orchestrator. This is currently a catalog-specific
```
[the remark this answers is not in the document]
scottmbaker: Can you be more specific? for the completion? Usually it's a `get` because it returns a blob of bash script (or similar) that is then executed in the shell.

design-proposals/orch-cli.md
scottmbaker: @Andrea-Campanella I think it may actually be part of the config file and you only need to use `-p` if the default is unset.
@SeanCondon I've avoided putting the list of nouns -- I feel like tenant-datamodel is a service like any other, and `list projects` and `create project` should be verb/noun combinations.

design-proposals/orch-cli.md
scottmbaker: moved.

design-proposals/orch-cli.md
scottmbaker: In my opinion, clusters/applications/ens managed by the orchestrator. @se-chris-thach is working on a proposal for an installer CLI (potentially to be combined here) and maybe it would make sense to also have operations on the components, to start/stop/get-status on them either in that CLI, or in a combined CLI.

design-proposals/orch-cli.md
scottmbaker: agreed, and updated.

design-proposals/orch-cli.md
scottmbaker: I discussed it briefly in the `subject` section and gave an example. But I think something better would be a good idea. @pierventre can you do a write-up based on EIM's filtering support? I know you have options for boolean operations, parenthesis, etc.

design-proposals/orch-cli.md
scottmbaker: Adding a paragraph to try to disambiguate the purpose.

design-proposals/orch-cli.md
scottmbaker: I have mixed feelings about this as if we're not careful, we could create things that have different names but turn out to be similar operations. For example, `create` and `register` do very similar things -- they both cause a new object to be created in the data model. I can see that `onboard` and `provision` are different, they modify the object that was previously created. I think we should leave this to our best judgement -- write up an ADR for the proposed EIM-specific verbs, and review.

design-proposals/orch-cli.md
scottmbaker: Using `kubectl` as an example, `kubectl delete pod <name> --force` is the way to force delete a pod. I feel like deauthorize could be implemented as a set/update operation. For example, `cli set host foo --deauthorized true`. Let's keep verb discussion in the previous thread.

design-proposals/orch-cli.md
scottmbaker: Some more thoughts -- using `kubectl` as an example, it supports the basic verbs (create, get delete, etc) but it also adds custom verbs where they make sense. For example, `kubectl cordon` applies to nodes and `kubectl port-forward` applies to apps and services. We can make similar accommodations where they make sense.

design-proposals/orch-cli.md
scottmbaker: agreed.

design-proposals/orch-cli.md
scottmbaker: These are called "application" in the GUI and the API (and the docs), so I feel like it would create a divergence in experience to call them something else. Ideally naming concepts should carry over between CLI and GUI. If we want to rename "application", we should take it up as a general app-orch discussion. We would need oversight from UX / Branding on renaming a core concept (it took some time to arrive at `Application` and `Deployment Package` so long ago)

design-proposals/orch-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+  - `cli update catalog application foo` ... catalog is a component name. It is not relevant to the user that
+    applications are managed by the catalog service.
+
+- The CLI should avoid distinguishing between singular and plural nouns. For example `list applications`
+  and `list application` should be equivalent commands.
+
+- The CLI stores its local state in the file (in a Linux environment) `~/.orch-cli/orch-cli.yaml`. This state
+  includes the orchestrator endpoint, keycloak endpoint, default projects, logged in username, and logged in
+  refresh token. This file should have permissions set appropriately.
```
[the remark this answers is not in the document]
scottmbaker: Currently I think we explicitly set the default project rather than last-used.

design-proposals/orch-cli.md
scottmbaker: My first thought would be `cli start <vm>` and `cli stop <vm>`. Grabbing the most similar example I can think of, `kubectl cordon node` and `kubectl uncordon node`. We could also fold it into update, such as `cli update <appname> -- running false`

design-proposals/orch-cli.md
scottmbaker: Currently the structure is to separate functionality into files, but I think as we get larger, adding subdirectories as a further organizational layer would make sense. As Sean mentioned, it's discussed below.

design-proposals/orch-cli.md
scottmbaker: Updated to use `get config` and `update config` verb/noun pairs.

scottmbaker:
> How about a subcommand like `cli install service aws`? This will trigger the install TUI on fresh installs.

I agree with @se-chris-thach that we should consider unifying these binaries, though I could use some clarification -- does the `cli install` command need to be run from inside the environment where it will be installed? or can it be run remotely? The CLI in general is intended to be run from an administrator laptop.

design-proposals/orch-cli.md
pierventre: It is very simple we adopt and probably catalog and other components adopt too this:
https://google.aip.dev/160

pierventre (approved):

damiankopyto (approved):

design-proposals/orch-cli.md
jokuniew: so according to this aip.dev/160 will it look like those examples
```sh
orch get --filter="name = 'item1'"
orch get --filter="status != 'pending'"
orch get --filter="name : 'item' AND status = 'ready'"
orch get --filter="(status = 'ready' OR status = 'pending') AND name != 'test'"
```
? @pierventre

scottmbaker: thanks -- I'm going to pull in filtering into a follow-on PR.

PR #250, opened 2025-04-30
design: Scale provisioning of EMT-S nodes

krishnajs:
### Description

This PR adds design proposal for Scale provisioning of EMT-S edge node supporting OXM workflow

N/A

### How Has This Been Tested?

N/A

### Checklist:


design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Proposal Summary
+
+EIM leverages the Tinkerbell solution for provisioning operating systems on edge nodes. The current implementation supports OS provisioning via bootable USB or iPXE/HTTPs boot. To enable scalable provisioning of EMT-S edge nodes for the OXM workflow, this proposal suggests integrating PXE-based provisioning into EIM by utilizing the `boots` component of Tinkerbell. Additionally, EMF and EIM will support configurations to deploy this as a standalone EIM solution (to be referred in future as EIM-S), tailored for OXMs to efficiently provision edge nodes at scale. OXM will have the option of provisioning edge nodes using bootable USB, iPXE/HTTPs boot, or PXE-based provisioning. The solution will also include a user experience (UX) for pre-registering edge nodes using serial numbers, UUIDs, or MAC addresses. Furthermore, the solution will support the provisioning of different operating system profiles based on the selected identifiers. In cases where a device on the local area network (LAN) boots over PXE and is not pre-registered, the default operating system will be provisioned
```
osinstom, line 13:
```suggestion
EIM leverages the Tinkerbell solution for provisioning operating systems on edge nodes. The current implementation supports OS provisioning via bootable USB or iPXE/HTTPs boot. To enable scalable provisioning of EMT-S edge nodes for the OXM workflow, this proposal suggests integrating PXE-based provisioning into EIM by utilizing the `boots` component of Tinkerbell. Additionally, EMF and EIM will support configurations to deploy the EMF as a standalone EIM solution (to be referred in future as EIM-S), tailored for OXMs to efficiently provision edge nodes at scale. OXM will have the option of provisioning edge nodes using bootable USB, iPXE/HTTPs boot, or PXE-based provisioning. The solution will also include a user experience (UX) for pre-registering edge nodes using serial numbers, UUIDs, or MAC addresses. Furthermore, the solution will support the provisioning of different operating system profiles based on the selected identifiers. In cases where a device on the local area network (LAN) boots over PXE and is not pre-registered, the default operating system will be provisioned
```

osinstom: > Furthermore, the solution will support the provisioning of different operating system profiles based on the selected identifiers.

Please clarify this - do you mean we can pre-register different ENs with different OS profiles?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,27 @@
...
+
+Author(s): EIM team.
+
+Last updated: 29.04.2025
+
+## Abstract
+
+The Edge Microvisor Toolkit Standalone (EMT-S) node is designed to enable enterprise customers and developers to evaluate Intel silicon-based platforms for Edge AI use cases. In this context, Original Equipment Manufacturers (OXMs) play a critical role by preparing edge nodes in bulk for end customers at their facilities before shipping them to the deployment locations. To support the OXM workflow for Edge AI use cases, it is essential to implement a scalable provisioning solution for multiple EMT-S nodes. This document outlines the design proposal for enabling large-scale provisioning of EMT-S edge nodes to meet the requirements of the OXM workflow.
+
+### Proposal Summary
```
osinstom, line 11: Should this proposal include what are the components of EIM-S that will be deployed on-prem? CC: @daniele-moro

osinstom: I believe we should not deploy infra-managers at all

design-proposals/emts-scale-provisioning.md
osinstom: > In cases where a device on the local area network (LAN) boots over PXE and is not pre-registered, the default operating system will be provisioned

Should we install default OS or just ignore any device that is not pre-registered? The latter seems more controlled environment to me.

design-proposals/emts-scale-provisioning.md
osinstom, line 11: We need more details on the workflow itself. This comes to my mind:
- what does boots require to perform provisioning? Do we need to extend the data model? Does it need MAC address? 
- how the interaction between Inventory and boots will look like? Do we need any additional component that coordinates the process or will it be a job of OM ?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,117 @@
...
+
+  ipxe->>+pa: Download boot.ipxe
+  pa->>-ipxe: [boot.ipxe]
+
+  note over bios,inv: OS provisioning continues in a standard flow, through HookOS and Tinkerbell workflow
+```
+
+Notes:
+- Similar to the standard flow - If EN is not pre-registered, the process will fallback to Interactive Onboarding waiting for a user to provide credentials.
+- Similar to the standard flow - If EN is pre-registered, but the OS profile is not selected, the default OS will be provisioned.
```
pierventre, line 97: this is not true - it will be stuck in onboarded state. We removed the auto-provisioning

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,117 @@
...
+The current design proposal allows to easily support legacy PXE boot while keeping the current UX around EN pre-registration. It also lets us to keep all the features that EIM currently supports,
+including logging, KPI instrumentation, observability, etc.
+
+The alternative considered was to use a standalone Tinkerbell deployment without the rest of the EIM stack. While it has clear advantage of a more lightweight deployment,
+it would completely change the UX as we would need to familiarize customers with Tinkerbell APIs (or have a custom CLI tool to help them manage Tinkerbell CRDs).
+With the current proposal we keep using the current UX, with possibility to use Bulk Import Tool to scale preregistration process and selectively choose OS profiles for Edge Nodes.
+
+## Open issues
+
+- By default, the provisioning flow will complete and all the agents will be installed, started and connected to the orchestrator instance that deployed them.
```
pierventre, line 114: another reason to target the graceful go-away logic as first priority too

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,117 @@
...
+### MVP requirements
+
+Following are the MVP requirements for the scale provisioning of EMT-S edge node supporting OXM workflow:
+
+- Provision multiple BareMetal edge nodes without onboarding for the purpose of standalone/singleton use.
+- Provide deploy a service on the local network that can achieve this provisioning at scale.
+- Deploy the provisioning service on the local network that support PXE Boot (BIOS/UEFI with DHCP + TFTP) boot and iPXE with HTTPs.  
+- Have a UX to pre-register BareMetal edge nodes using Serial number or UUID or MAC address.
+- Provision different OS profiles to different edge nodes selected based on Serial number or UUID or MAC address.
+- Provision default OS when a device on the LAN boots over PXE and is not pre-registered.
```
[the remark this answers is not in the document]
osinstom: If we want to meet this requirement, we should re-enable auto-provisioning. IMO we can use the current behavior and stop at HookOS stage.

design-proposals/emts-scale-provisioning.md
daniele-moro: IMO, yes, we should clarify which components of EIM we really need to deploy. I guess API, Inventory, OM, TC (?), DKAM (?).
We may not care about multi-tenancy for this use case, but we are currently tightly coupled with TC+OSRM for tenant bootstrap, in order for EMF-EIM to work. But we could figure out a way to overcome this limitation.
Also, we are currently tightly coupled with multiple platform provided components (DB, Keycloak, Vault...).

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,142 @@
...
+
+The solution assumes that we will deploy a slimmed down version of EIM (aka EIM standalone) on a customer's premises (OXM warehouse). The slimmed-down EIM
+should only consist of required components to drive OS provisioning.
+
+### Slimmed-down EIM
+
+The provisioning of EMT-S at scale will be driven by a local orchestrator instance that will be slimmed down to include only necessary components. In a nutshell,
+it will consist of:
+
+- Foundational Platform Services that are required to deploy and run the orchestrator instance.
```
daniele-moro, line 39: Argo? How do we handle the deployment of the orchestrator?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,142 @@
...
+- MetalLB services
+
+The following FPS/Observability components should be disabled:
+- Kyverno
+- Prometheus and all metrics-related components (including infra-core's exporter, Mimir)
+- SRE exporter
+- Alerting monitors
+- Loki should be scaled down to minimal deployment
+
+**The installation of EIM-S should become a one-line command operation.**
```
daniele-moro, line 61: What are the assumption? A VM? A baremetal machine? a k8s cluster?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Foundational Platform Services that are required to deploy and run the orchestrator instance.
+- Limited observability stack that should only provide logs.
+- A reduced flavor of EIM - only infra-core and infra-onboarding will be deployed (no infra-managers as they won't be used by EMT-S).
+- Limited Web UI with EIM only.
+- No cluster and application orchestrator deployed.
+- Abandon HA requirements - EIM-S is primarily a single-node, on-prem deployment. Any data loss can be backed by hardware-level redundancy. Think of EIM-S as Rancher Desktop or Virtualbox.
+
+FPS components that are essential to EIM are:
+- IAM, Multi-Tenancy components
```
daniele-moro, line 47: Is MT useful for this use case? If not, we should provision the single org/project and users in the provisioning phase to avoid the burden of bootstrapping tenants and users.

design-proposals/emts-scale-provisioning.md
daniele-moro: @pierventre autoprovision is not removed, but only disabled. We could re-enable it, if we think it's useful for this use case. @krishnajs worth mentioning this, since by default it would be disabled.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,142 @@
...
+EIM stack.
+
+By default, Tinkerbell SMEE relies on MAC addresses to uniquely identify PXE-booting machines and customize the iPXE script per machine.
+In the case of EIM, we use a static iPXE script that is not customized per Edge Node. Therefore, we can avoid using MAC address as a unique identifier
+and it will let use avoid adding MAC address as another EN identifier during pre-registration.
+
+The below flow will be feasible to achieve with minimal modifications to the EIM stack. The following changes must be made:
+
+- Tinkerbell SMEE must be enabled via infra-charts to provide DHCP and TFTP server
+- Tinkerbell SMEE must be configured with the following flags to avoid lookups by MAC address and to use our EIM iPXE script:
```
daniele-moro, line 75: worth clarifying that we don't have requirements on local network DHCP server.

design-proposals/emts-scale-provisioning.md
pierventre: I know that is not removed but this is something that people did not like as UX and we disabled because People were not happy about.

daniele-moro: Yes, also we can achieve an equivalent workflow even with autoprovision disabled.

design-proposals/emts-scale-provisioning.md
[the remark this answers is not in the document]
Andrea-Campanella: IMHO if it's needed we can re-enable but do not advertise, this scenario was the reason why we had auto-provision in the first place.

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: yes, let's assume for now this is deployed via the EIM-Standalone profile you are working on @daniele-moro so effectively we leverage the orchestrator on prem installer.

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: yes, we can make this assumption, let's include it in the EIM standalone profile @daniele-moro

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: let's go with the assumption we use the orch-installe for now and avoid going too deep into this optimization

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,142 @@
...
+## Affected components and Teams
+
+- **FPS team** should work with the help of EIM team on slimming down the FPS components to meet EIM-S requirements.
+- **EIM team** responsible for slimming down deployment of EIM and enabling Tinkerbell SMEE with required configuration. 
+
+## Open issues
+
+- By default, the provisioning flow will complete and all the agents will be installed, started and connected to the orchestrator instance that deployed them.
+  We may need a way to stop the provisioning flow without installing and starting agents.
+- In this solution, PXE boot will always be allowed for any device in a local subnet trying to initiate PXE boot.
```
Andrea-Campanella, line 140: I'm ok with this assumption, we assume the OXM takes security of his network and also that it has a know list of devices trying to PXE boot

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: agreed.

design-proposals/emts-scale-provisioning.md
se-chris-thach: I strongly advise against creating a dedicated SKU for this, like we previously did in a previous release. 

In a parallel work effort, I proposed that we make Orchestrator configurable at initial deployment time AND runtime. 

In this effort, I propose that instead of a separate variant of Orchestrator, we utilize this proposed dynamically configurable Orchestrator and explicitly configure it to disable unnecessary features).

cc @charlesmcchan

se-chris-thach (commented):
Can we please add detail around the security considerations for this new design?

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: yes, no variant is intended, a profile is what we are going to use for now, further optimizations can be achieved in the future.

charlesmcchan: There is already a way to do that via Argo CD profile, but it is not as straightforward as it can be. I will make "the ability to enable/disable components easily" an explicit requirement for the new installer that we are working on.

design-proposals/emts-scale-provisioning.md
osinstom: Clarified above in the new version

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+  - MetalLB
+- Relaxed security requirements:
+  - No Istio for inter-Pod communication security
+  - No Kyverno for policies
+- No UI deployed, the EN pre-registration is done via Bulk Import Tool, API or CLI tool.
+- Limited observability stack that should only provide logs. No alerting, SRE, Prometheus.
+- Abandon HA requirements - EIM-S is primarily a single-node, on-prem deployment. Any data loss can be backed by hardware-level redundancy. Think of EIM-S as Rancher Desktop or Virtualbox.
+
+The minimal EIM deployment should consist of:
+- infra-core without exporter and
```
daniele-moro, line 292:
```suggestion
```

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+The alternative considered was to use a standalone Tinkerbell deployment without the rest of the EIM stack. While it has advantage of a more lightweight deployment,
+it would completely change the UX as we would need to familiarize customers with Tinkerbell APIs (or have a custom CLI tool to help them manage Tinkerbell CRDs).
+With the current proposal we keep using the current UX, with possibility to use Bulk Import Tool to scale preregistration process and selectively choose OS profiles for Edge Nodes.
+
+## Affected components and Teams
+
+**FPS team** needs provide help for the following:
+- Reduce FPS components to minimum to make EIM-S deployment lightweight
+- Make EIM-S deployment operation easy in terms of number of CLI commands/steps to execute
```
daniele-moro, line 307: I would also add "fast", for example checking if we can parallelize anything that we currently don't parallelize.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+Following are the MVP requirements for the scale provisioning of EMT-S edge node supporting OXM workflow:
+
+- Provision multiple BareMetal edge nodes without onboarding for the purpose of standalone/singleton use.
+- Provide deploy a service on the local network that can achieve this provisioning at scale.
+- Deploy the provisioning service on the local network that support PXE Boot (BIOS/UEFI with DHCP + TFTP) boot and iPXE with HTTPs.  
+- Have a UX to pre-register BareMetal edge nodes using Serial number or UUID or MAC address.
+- Provision different OS profiles to different edge nodes selected based on Serial number or UUID or MAC address.
+- Provision default OS when a device on the LAN boots over PXE and is not pre-registered.
+- Have a ux of  collecting provisioning logs and status of edge nodes.
```
Ram-srini, line 25: Should this ADR also cover the A/B upgrade for Day1 flows? We can't use the USB based A/B upgrade, because it is not scalable.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Proposal Summary
+
+EIM leverages the Tinkerbell solution for provisioning operating systems on edge nodes. The current implementation supports OS provisioning via bootable USB or iPXE/HTTPs boot. To enable scalable provisioning of EMT-S edge nodes for the OXM workflow, this proposal suggests integrating PXE-based provisioning into EIM by utilizing the `smee` (formerly `boots`) component of Tinkerbell. Additionally, EMF and EIM will support configurations to deploy this as a standalone EIM solution (to be referred in future as EIM-S), tailored for OXMs to efficiently provision edge nodes at scale. OXM will have the option of provisioning edge nodes using bootable USB, iPXE/HTTPs boot, or PXE-based provisioning. The solution will also include a user experience (UX) for pre-registering edge nodes using serial numbers, UUIDs, or MAC addresses. Furthermore, the solution will support the provisioning of different operating system profiles based on the selected identifiers. In cases where a device on the local area network (LAN) boots over PXE and is not pre-registered, the default operating system will be provisioned
```
Ram-srini, line 13: Non-interactive way of onboarding with pre-registration doesn't make sense to PXE/TFTP based flow, because ENs within the local networks are get discovered with DHCP and get the PXE EFI app over TFTP both are layer 2 protocols. Is customer restrict few EN based on the pre-registration? In my opinion pre-registration is overhead for the customer.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+### MVP requirements
+
+Following are the MVP requirements for the scale provisioning of EMT-S edge node supporting OXM workflow:
+
+- Provision multiple BareMetal edge nodes without onboarding for the purpose of standalone/singleton use.
+- Provide deploy a service on the local network that can achieve this provisioning at scale.
+- Deploy the provisioning service on the local network that support PXE Boot (BIOS/UEFI with DHCP + TFTP) boot and iPXE with HTTPs.
```
Ram-srini, line 21: We need to call out, how the PXE/TFTP provisioned EN gets attached to full blown EMF orchestrator(provisioning, app orch, cluster orch).

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+### MVP requirements
+
+Following are the MVP requirements for the scale provisioning of EMT-S edge node supporting OXM workflow:
+
+- Provision multiple BareMetal edge nodes without onboarding for the purpose of standalone/singleton use.
+- Provide deploy a service on the local network that can achieve this provisioning at scale.
+- Deploy the provisioning service on the local network that support PXE Boot (BIOS/UEFI with DHCP + TFTP) boot and iPXE with HTTPs.  
+- Have a UX to pre-register BareMetal edge nodes using Serial number or UUID or MAC address.
```
Ram-srini, line 22: It should be optional and not mandatory.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+- Provision different OS profiles to different edge nodes selected based on Serial number or UUID or MAC address.
+- Provision default OS when a device on the LAN boots over PXE and is not pre-registered.
```
Ram-srini, line 24: What is the meaning of default OS here, Are we installing supporting the Ubuntu also in EMT-S?

design-proposals/emts-scale-provisioning.md
Andrea-Campanella: let's have this as a separate ADR, for the scope of the e2e provisioning it does not matter if we have 1 or 100 components and FPS elements.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+The iPXE binary provided by the SMEE's TFTP server is a third-party script, maintained by the Tinkerbell project.
+
+The solution assumes that the local EIM-provided DHCP/TFTP server is only used to initiate boot via legacy PXE (as legacy PXE has limited capabilities, i.e., it doesn't support HTTPS boot),
+but, once chain-loaded to iPXE, the Micro-OS is downloaded and used to drive provisioning.
+In other words, the only difference between the new PXE-based boot and HTTP-based boot is how the OS provisioning is triggered. The subsequent workflow remains the same -
+it leverages Micro-OS, device discovery and Tinkerbell workflow to complete OS provisioning.
+
+**NOTE1**: Customers should provide their own local DHCP server for dynamic IP address assignment.
+
+**NOTE2**: The workflow assumes that the EIM Standalone is deployed locally on customers' premises and ENs have direct connectivity with EIM services.
```
pierventre, line 52: shall we say something about the interaction with external (not managed by us) DHCP servers?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+The high-level PXE-based provisioning workflow is as follows:
```
pierventre, line 54: also in the sequence say something about external DHCP interactions. Or create a small paragraph with the assumptions and the checklist user should comply:

- Boot works in proxy dhcp mode only
- External DHCP server uses DHCP options....

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+
+Currently, Provisioning Nginx is deployed behind HTTPS and the iPXE binary must be built with orchestrator's CA certificate for successful handshake.
+Tinkerbell SMEE obviously doesn't include the EMF's CA certificate. Therefore, any communication over HTTPS is not feasible.
+
+For on-prem, EIM standalone profile only we will disable HTTPS for Provisioning Nginx. The assumption is that the on-prem, customers' sites are typically
+controlled environments and we can relax security requirements to support PXE boot (any traffic doesn't leave a local L2 network).
+
+3. **Expose Tinkerbell SMEE's DHCP/TFTP server via K8s External IP**
+
+Tinkerbell SMEE's DHCP/TFTP server must be reachable from a local, on-prem L2 network to handle DHCP/TFTP requests. This requires modifications to FPS services.
```
pierventre, line 145: is it not enough use hostNetwork for port 67-69? Which modifications do we expect?

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+The EIM-local assists in the initial PXE boot and make possible for ENs to initiate boot via PXE. Once booted into Micro-OS, the provisioning process is taken over by the cloud-based orchestrator.
+
+The EIM-local consists of the following components:
+1. **Standalone Tinkerbell SMEE** providing DHCP/TFTP server to support legacy PXE boot.
+2. **Local HTTP server** (e..g, Nginx) storing mirrors of `boot.ipxe` and Micro-OS image.
+3. (OPTIONAL) **K8s cluster with MetalLB extension** to make Standalone SMEE's DHCP/TFTP servers accessible from a local network. Only needed if EIM-local is deployed on top of Kubernetes.
+   Note that the EIM-local can also be deployed as standalone OS services or Docker containers with `--network=host`.
+
+**NOTE1:** Local HTTP server providing `boot.ipxe` and Micro-OS image is needed to overcome HTTPS issue as
+SMEE's built-in iPXE doesn't include EMF's CA certificate. This alternative design assumes no modifications to Tinkerbell SMEE, for simplicity.
```
pierventre, line 162: Honestly, I argue that improving DKAM seems to be a better option here. Of course I prefer the 2nd option below

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+  - Vault (?)
+  - Traefik
+  - cert-manager
+  - MetalLB
+- Relaxed security requirements:
+  - No Istio for inter-Pod communication security
+  - No Kyverno for policies
+- No UI deployed, the EN pre-registration is done via Bulk Import Tool, API or CLI tool.
+- Limited observability stack that should only provide logs. No alerting, SRE, Prometheus.
+- Abandon HA requirements - EIM-S is primarily a single-node, on-prem deployment. Any data loss can be backed by hardware-level redundancy. Think of EIM-S as Rancher Desktop or Virtualbox.
```
pierventre, line 289: we dont have any HA mechanism when deployed locally

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+- Tinkerbell SMEE must be enabled via infra-charts to provide DHCP and TFTP server
+- Tinkerbell SMEE must be configured with the following flags to avoid lookups by MAC address and to use our EIM iPXE script:
+  - `-dhcp-mode=auto-proxy`
+  - `-dhcp-http-ipxe-script-url=<URL to iPXE script on Provisioning NGINX>` (e.g., `http://tinkerbell-nginx.CLUSTER_DOMAIN/tink-stack/boot.ipxe`)
+  - `-dhcp-http-ipxe-script-prepend-mac=false`
+
+2. **Expose Provisioning Nginx via HTTP for on-prem EIM standalone only**
+
+Currently, Provisioning Nginx is deployed behind HTTPS and the iPXE binary must be built with orchestrator's CA certificate for successful handshake.
+Tinkerbell SMEE obviously doesn't include the EMF's CA certificate. Therefore, any communication over HTTPS is not feasible.
```
Ram-srini, line 138: Good to check with SAFE team once to go ahead with HTTP provided local network is protected with firewall.

design-proposals/emts-scale-provisioning.md
osinstom: Resolving the comment, we will use new on-prem installer being designed in https://github.com/open-edge-platform/edge-manageability-framework/pull/290.

design-proposals/emts-scale-provisioning.md
osinstom: I'll add a diagram with a sample network topology but we're not really making any changes to network requirements. We had been relying on customer-managed DHCP server for long time already

design-proposals/emts-scale-provisioning.md
osinstom: Perhaps, I'll work on another PoC to make SMEE work on the on-prem deployment

design-proposals/emts-scale-provisioning.md
osinstom: Even in the observability stack ?

design-proposals/emts-scale-provisioning.md
pierventre: I believe we need to tell the user to change few things in their dhcp server too? Dont we need the next-server option?

osinstom: next-server is provided by SMEE's DHCP server which is configured in proxy DHCP mode

osinstom: If we enforce any change to DHCP servers, this design proposal wouldn't make sense to me

pierventre: ok this is what I found to make them coexist:
- A PXE client sends a DHCP request (broadcast).
- The external DHCP server responds with: An IP address, The next-server (Boots server IP), The filename (iPXE binary).
- Tinkerbell SMEE (in proxy DHCP mode) responds with PXE-specific options (e.g., boot filename and TFTP/HTTP server details).
- The PXE client downloads the iPXE binary and executes the boot script provided by Boots.

osinstom: let me do some more experiments to confirm once I have my target on-prem environment running

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,245 @@
...
+We won't be there yet for 3.1, but this should be a direction for us, so that we don't have diverged codebases.
+
+### EIM-only profile of EMF deployment (EIM-S)
+
+The provisioning of EMT-S at scale will be driven by a local orchestrator instance that only includes EIM components. 
+The main requirement of EIM standalone is to make it easy and resource-efficient for customers to deploy.
+
+For MVP, we will leverage the EIM Standalone profile. This is an EMF deployment profile that only deploys EIM with Observability and FPS components.
+
+EIM Standalone profile for minimal OXM wareshouse deployments will be configured with EMT-Standalone OS profile only.
```
daniele-moro, line 202:
```suggestion
EIM Standalone profile for minimal OXM warehouse deployments will be configured with EMT-Standalone OS profile only.
```

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,245 @@
...
+### MVP requirements
+
+Following are the MVP requirements for the scale provisioning of EMT-S edge node supporting OXM workflow:
+
+- Provision multiple BareMetal edge nodes without onboarding for the purpose of standalone/singleton use.
+- Provide deploy a service on the local network that can achieve this provisioning at scale.
+- Deploy the provisioning service on the local network that support PXE Boot (BIOS/UEFI with DHCP + TFTP) boot and iPXE with HTTPs.  
+- Have a UX to pre-register BareMetal edge nodes using Serial number or UUID.
+- Provision different OS profiles to different edge nodes selected based on Serial number or UUID.
+- Provision default OS when a device on the LAN boots over PXE and is not pre-registered.
```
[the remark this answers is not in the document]
osinstom: @krishnajs is this requirements still valid with pre-registration flow? I think we should remove it.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,255 @@
...
+
+  ipxe->>+pa: Download Micro-OS
+  pa->>-ipxe: [Micro-OS image]
+
+  note over ipxe,microos: Boots into Micro-OS
+
+  note over microos,inv: OS provisioning continues with the standard flow, through Micro-OS and Tinkerbell workflow
+```
+
+1. Users perform standard EN registration via UI or Bulk Import Tool. For each EN we can define OS profile and additonal settings (e.g., site, local account).
```
teone, line 105: > Users perform standard EN registration via UI

The registration flow is not the standard one as whether the Cluster is installed or not entirely depend on the selection of the OS profile (via cloud-init). We should make sure that the user only has the `<project_id>_ir/irw` roles so that the UI can hide the Cluster creation steps from the registration/provisioning flow.

cc @satya-in @soniabha-intc

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,255 @@
...
+
+> Note: It might be possible for EMF-EIM to support provisioning of the EMT-S nodes. supporting this capability as part of MVP depends on any active customer requirements.
+
+## Solution
+
+The solution consists of three major enhancements (modifications) to EMF:
+1) **Support legacy PXE boot to scale EMT-S provisioning** - the EIM will be extended with a local DHCP/TFTP server that helps initiate OS provisioning via legacy PXE boot.
+   The legacy PXE boot uses the devices' PXE firmware to bootstrap into iPXE, which then continues the EN provisioning process.
+2) **Extend usage of Platform Bundle to be compatible with USB-based EMT-S** - to be consistent with USB-based EMT-S, the EIM will consume Platform Bundle that contains all the scripts and files to install standalone K8s cluster and other customizations.
+3) **Use EIM standalone deployment** - the proposed solution involves deploying a specific EMF profile, referred to as EIM standalone profile.
```
Andrea-Campanella, line 36: In 3.1 we deploy the whole orchestrator in the OXM flow reducing the resource requirements (TODO: measure resources needed for 100 within 5 minutes, also unlimited resources as good as we can get numbers.) and use roles NOT to show App and CO elements in UI, default MT user, projects and org created automatically via a module during installation, with a PoC of Infra-only profile of EMF with "extreme" simplification but still using the orchestrator as a whole, not part of 3.1. Reason behind decision to keep the whole of EMF is limited resource/performance/time and UX gain compared to the effort we need to invest to maintain. PoC of Daniele will be revisited for a possible 3.2 implementation.

as per the discussion please move this lower in the ADR and advertise as PoC.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,255 @@
...
+admission control to prevent unwanted ENs being provisioned to EMF "magically". 
+
+The alternative considered was to use a standalone Tinkerbell deployment without the rest of the EIM stack. While it has advantage of a more lightweight deployment,
+it would completely change the UX as we would need to familiarize customers with Tinkerbell APIs (or have a custom CLI tool to help them manage Tinkerbell CRDs).
+With the current proposal we keep using the current UX, with possibility to use Bulk Import Tool to scale preregistration process and selectively choose OS profiles for Edge Nodes.
+
+Also, an alternative to using SMEE was considered to avoid divergence in security policies (HTTP vs. HTTPS for Provisioning Nginx).
+This direction requires more development effort, but is still left as future improvement. It's further elaborated in a separate [design proposal](https://github.com/open-edge-platform/edge-manageability-framework/pull/309).
+
+## Affected components and Teams
```
Andrea-Campanella, line 230: @daniele-moro to add the role to avoid User accessing App and CO.

design-proposals/emts-scale-provisioning.md
osinstom: After some experiments, the decision is to use off-the-shelf DHCP server to provide "our" `signed_ipxe.efi`, see PoC charts: https://github.com/open-edge-platform/infra-charts/pull/53

design-proposals/emts-scale-provisioning.md
daniele-moro: We can use this group https://docs.openedgeplatform.intel.com/edge-manage-docs/main/shared/shared_iam_groups.html#project-id-host-manager-group, as long as `[<project-id>_im-rw](https://docs.openedgeplatform.intel.com/edge-manage-docs/main/shared/shared_iam_groups.html#project-id-im-rw)` role is there, we are ok from edgeinfra perspective.

design-proposals/emts-scale-provisioning.md
```diff
@@ -0,0 +1,321 @@
...
+The solution relies on the capabilities provided by Tinkerbell SMEE. SMEE provides a minimal DHCP and TFTP server implementations that support PXE boot.
+The DHCP server provides necessary PXE info to machines initiating the PXE boot, while the TFTP server provides the iPXE binary that can be chain-loaded at boot time by PXE.
+The iPXE binary provided by the SMEE's TFTP server is a third-party script, maintained by the Tinkerbell project.
+
+The solution assumes that the local EIM-provided DHCP/TFTP server is only used to initiate boot via legacy PXE (as legacy PXE has limited capabilities, i.e., it doesn't support HTTPS boot),
+but, once chain-loaded to iPXE, the Micro-OS is downloaded and used to drive provisioning.
+In other words, the only difference between the new PXE-based boot and HTTP-based boot is how the OS provisioning is triggered. The subsequent workflow remains the same -
+it leverages Micro-OS, device discovery and Tinkerbell workflow to complete OS provisioning.
+
+**NOTE1**: Customers should provide their own local DHCP server for dynamic IP address assignment.
```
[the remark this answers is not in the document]
osinstom: clarified

design-proposals/emts-scale-provisioning.md
osinstom: after experiments, I've updated the proposal. Also added network diagram to clarify network reqs

design-proposals/emts-scale-provisioning.md
osinstom: done

Andrea-Campanella (approved):

design-proposals/emts-scale-provisioning.md
```diff
@@ -192,20 +182,28 @@ that is almost equal to Tinker action doing the same.
...
-### EIM-only profile of EMF deployment (EIM-S)
+### Detach of Edge Nodes
 
-The provisioning of EMT-S at scale will be driven by a local orchestrator instance that only includes EIM components. 
-The main requirement of EIM standalone is to make it easy and resource-efficient for customers to deploy.
-
-For MVP, we will leverage the EIM Standalone profile. This is an EMF deployment profile that only deploys EIM with Observability and FPS components.
+Once ENs are provisioned and ready to be shipped to the field to operate as standalone ENs, they should be detached from the local EMF orchestrator.
+An explicit detach would be needed if there are agents' configurations, certificates, credentials, etc. installed on ENs.
+However, **BM agents won't be configured and activated for EMT-S OS profiles**. Therefore, the only operation to detach ENs from the orchestrator is to delete them via northbound API.
```
zdw, line 189: On the orchestrator side, we should mention how this interacts with revocation of the auth credentials in Keycloak when an EN is removed, or point to docs that talk about this.

Andrea-Campanella (approved):

PR #251, opened 2025-04-30
adr: cluster-orch cli

jokuniew:
### Description

Introduce nouns for cluster orchestration part of cli

none.

### How Has This Been Tested?
none.

### Checklist:


design-proposals/co-cli.md
```diff
@@ -0,0 +1,101 @@
...
+https://github.com/open-edge-platform/edge-manageability-framework/pull/246.
+
+## Proposal
+
+The CLI follows a `verb - noun - subject` pattern.
+
+The target for cluster-orchestration part in CLI is to be able to:
+- create/list/delete clustertemplates
+- mark clustertemplate as default one
+- create cluster with given hosts and given template (or pick default template if not specified)
```
gcgirish, line 22: Although not specifically mentioned here, I believe delete cluster should also be supported

design-proposals/co-cli.md
```diff
@@ -0,0 +1,101 @@
...
+- `clustertemplate`
+- `cluster`
+- `en-kubeconfig`
+- `hw-features` (?)
+
+#### Syntax
+Assumption: All CLI commands should include a `--project` or `-p` parameter.
+For simplicity, this parameter is omitted in the commands provided below.
+
+##### clustertemplate
```
gcgirish, line 36: Do you have an example here on how filtering based commands would look like?

design-proposals/co-cli.md
```diff
@@ -0,0 +1,101 @@
...
+Cluster Orchestration
+
+## Implementation plan
+
+TODO:
+[A description of the implementation plan, who will do them, and when.
+This should include a discussion of how the work fits into the product's
+quarterly release cycle.]
+
+## Open issues (if applicable)
```
gcgirish, line 93: Is the option to display the version of CO components being used considered? Is that part of the larger scope of the CLI?

jokuniew: I believe the assumption was that we gather info only from API. I haven't seen such info in our apis so i would say, it's not possible.. 

if we can access kubectl with CLI, it would be doable

design-proposals/co-cli.md
jokuniew: Yes, info about get/delete added here 
https://github.com/open-edge-platform/edge-manageability-framework/pull/251/commits/006be86051c530e04db93ed5745c2fdfaea778d4

gcgirish (approved):

design-proposals/co-cli.md
```diff
@@ -0,0 +1,99 @@
...
+cli get cluster details <nodeId>
+cli apply cluster --file <path_to_cluster.yaml>
+cli create cluster \
+    --name <name> \
+    --hosts 6e6422c3-625e-507a-bc8a-bd2330e07e7e:all \ # required in format <uuid:role>
+    --labels key:value \ # optional
+    --template <template name-version> # optional
+cli delete cluster <name>
+cli delete cluster <name> --force
+cli set/update cluster --labels key:value, key2:value2
```
Andrea-Campanella, line 70: is it set or update ?

jokuniew: according to general adr - both 
https://github.com/open-edge-platform/edge-manageability-framework/pull/246/files#diff-c189904586748e35b97654c9134bf6fe3bdfbf0d9b141fa942678dd9f47e4ecaR71
image

design-proposals/co-cli.md
```diff
@@ -0,0 +1,99 @@
...
+cli get clustertemplate versions <name>
+cli apply clustertemplate --file <path_to_clustertemplate.yaml>
+cli delete clustertemplate <name> --version <version>
+```
+Cluster template does not support set/update operation.
+It is immutable.
+
+Extra operations related to clustertemplate:
+```bash
+cli set-default clustertemplate <name> --version <version>
```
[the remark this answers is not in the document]
jokuniew: default can be applied after cluster creation, so i'll keep the `set-default` verb
I'll add `--default` flag to `apply`

PR #268, opened 2025-05-01
Add skip download when using locally built artifacts

palade:
### Description

When setting use_local_build_artifact to true, the resource write_installer_config is not executed. This results in an error when set_proxy_config resource is run because it needs the artifact archive to be extracted so that it can install the proxy configuration. 

As part of this PR, the write_installer_config resource is updated to run only when enable_auto_install is set and to skip download if use_local_build_artifact is set to true.

N/A

### How Has This Been Tested?

Locally.

### Checklist:


emmakenny (approved):

se-chris-thach (approved):
Thanks so much for putting such a well-written description. LGTM!

PR #280, opened 2025-05-07
CI to test onprem Upgrades 

garyloug:
Initial GitHub Action to test On-Prem upgrade.
Will need to be updated with command when upgrade is functional.

### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


johnoloughlin (approved):

.github/workflows/upgrade.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  upgrade-from-version:
+    name: Upgrade from version
+    strategy:
+      matrix:
+        # Add comma separated Git tags here to test upgrade from different versions, e.g. [2.0.0, v3.0.0]
+        version-tags: [v3.0.0]
```
palade, line 22: Maybe these can be provided as input? e.g., 

inputs:

palade (approved):
LGTM, I left one comment

.github/workflows/upgrade.yml
garyloug: Wasn't aware. We have more changes to follow, so I'll look into inputs.

PR #285, opened 2025-05-08
Update gitea credential namespace

palade:
### Description

The `argocd-gitea-credential` is created under the `orch-platform` namespace, but on the onprem, this is created under `gitea`. This should be changed to `gitea` in the deploy target to match onprem.

none.

### How Has This Been Tested?

in the ci.

### Checklist:


palade:
Closing this for now as a different approach is required to tackle these changes in the installer.

PR #297, opened 2025-05-10
Fix GHA condition to ensure a job is triggered only after dependent jobs are successful

ashridatta:
### Description

Fix condition in following jobs to ensure these jobs are triggered only when dependent jobs are successful
- build and publish
- deploy kind
- deploy on-prem

Resolves issue where CI jobs continued to execute when version-check failed, which is not intended

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


charlesmcchan (approved):
The change itself looks good. But I am wondering why we have `always` in the first place?
Do we really want to spend the CI resource running kind and on-prem VIP when a PR doesn't even pass the basic version checks?
It might make sense to continue even after failure when the target is time consuming. We want it to finish and report all the problems at once. 
But when the target is small such as version check, the cost to re-run it later is much smaller.
I remember @teone  has made a comment in this area. Just curious what's your take Teo?

ashridatta:
> The change itself looks good. But I am wondering why we have `always` in the first place? Do we really want to spend the CI resource running kind and on-prem VIP when a PR doesn't even pass the basic version checks? It might make sense to continue even after failure when the target is time consuming. We want it to finish and report all the problems at once. But when the target is small such as version check, the cost to re-run it later is much smaller. I remember @teone has made a comment in this area. Just curious what's your take Teo?

Thanks Charles. I feel having always() for build-publish and deploy-* should be avoided as we dont want these to run if any previous one fail. However, I am assuming the reason behind executing these steps to report all error at ones. May be @se-chris-thach can confirm.

with version-check guard rail, the time consuming jobs should not be triggered. I feel it should be ok to run the rest of them as they wont take much of CI resources

se-chris-thach (approved):

PR #368, opened 2025-05-28
Disable DMT 

DevipriyaS17:
### Description

Disabling AMT helm chart for now to unblock the pipeline and will re-enable it later once we have fix.

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/argocd/applications/configs/infra-external.yaml
+++ b/argocd/applications/configs/infra-external.yaml
@@ -5,7 +5,7 @@
 ---
 import:
   amt:
-    enabled: false #Will Enable when ITEP-66997 is addressed
+    enabled: false # Will Enable when ITEP-66997 is addressed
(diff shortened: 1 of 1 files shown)


argocd/applications/configs/infra-external.yaml
```diff
@@ -5,7 +5,7 @@
 ---
 import:
   amt:
-    enabled: true
+    enabled: false
```
pierventre, line 8: `Enabled when ITEP-66997 is addressed`

pierventre: please add before merging

DevipriyaS17: Added the comment in the file so we keep track of it.

PR #392, opened 2025-05-31
feat(config): multiple improvements

charlesmcchan:
### Description

- Add configs (EN proxy, EKS IAM role, jump host IP, jump host key)
- Add README for installer
- Fix config loading for comma-separated fields
- Move TargetLables from user config to runtime state. Read it from installer CLI flag
- Improve unit test coverage

N/A

### How Has This Been Tested?

Unit test

### Checklist:

- [X] I agree to use the APACHE-2.0 license for my code changes
- [X] I have not introduced any 3rd party dependency changes
- [X] I have performed a self-review of my code


charlesmcchan:
I am going to skip VIP and merge this since VIP doesn't add any value to this PR.
There is a warning from gitleaks. Those are all dummy certs I created in the unit tests. They were excluded before but I guess the line number changes so we need to updated the baseline again. 

@cgoea any objection if I exclude the entire go test file from the gitleaks? As we keep adding unit tests, I expect this file will trigger the warning often.

PR #426, opened 2025-06-03
Create reusable GHA to deploy KIND

teone:
### Description

Provides a reusable github action to deploy the Kind orchestrator on a given commit ID so that other teams can leverage it for deployment.

### How Has This Been Tested?

In Progress

### Checklist:


.github/actions/deploy_kind/action.yaml
```diff
@@ -0,0 +1,78 @@
...
+      GIT_TOKEN: ${{ inputs.token }}
+      EDGE_MANAGEABILITY_FRAMEWORK_REV: ${{ inputs.orch_version }}
+      DOCKERHUB_TOKEN: ${{ inputs.docker_password }}
+      DOCKERHUB_USERNAME: ${{ inputs.docker_username }}
+      ORCH_DEFAULT_PASSWORD: ${{ inputs.orch_password }}
+    shell: bash
+    run: |
+      set +x
+      echo "Starting to deploy Orchestrator..."
+      mage deploy:kindAll
```
adibrastegarnia, line 58: Can we add a flag to deploy minimal or kindAll depending on the need? Something like this 

if [ "${{ inputs.deployment_type }}" == "minimal" ]; then
        echo "Using minimal deployment configuration"
        mage deploy:kindMinimal
      else
        echo "Using full deployment configuration"
        mage deploy:kindAll

teone: Yeah that's on my plan, but want to get it to work first :)

yi-tseng-intel (approved):

PR #466, opened 2025-06-11
feat(installer): add pull through cache proxy step

yi-tseng-intel:
### Description

Adds pull through cache proxy for new installer

N/A

### How Has This Been Tested?

Unit tests for step and Terraform module test

### Checklist:


new-installer/targets/aws/iac/utils/utils.go
```diff
@@ -33,6 +33,7 @@ import (
 const (
 	DefaultTestRegion         = "us-west-2"
 	DefaultTestCustomerTag    = "test-customer"
+	DefaultTestRoute53Zone    = "iteptest.infra-host.com"
```
charlesmcchan, line 36: This only works for our account.
Is there a way that community user can check this out and still make it work in their own aws account?

PR #473, opened 2025-06-12
feat: implement RDS step

charlesmcchan:
### Description

- Implement RDS step
- Fix cluster name randomization to avoid conflict between runs
- Address code scanning concern #520

N/A

### How Has This Been Tested?

Unit test

### Checklist:

### DIFF
--- a/new-installer/targets/aws/iac/rds/rds_test.go
+++ b/new-installer/targets/aws/iac/rds/rds_test.go
@@ -120,4 +120,7 @@ func (s *RDSTestSuite) TestApplyingModule() {
 
 	password := terraform.Output(s.T(), terraformOptions, "password")
 	s.NotEmpty(password, "Expected RDS Password to be created")
+
+	passwordID := terraform.Output(s.T(), terraformOptions, "password_id")
+	s.NotEmpty(passwordID, "Expected RDS Password ID to be created")
 }


new-installer/internal/steps/aws/rds.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+		}
+		if password, ok := terraformStepOutput.Output["password"]; ok {
+			runtimeState.Database.Password = strings.Trim(string(password.Value), "\"")
+		} else {
+			return runtimeState, &internal.OrchInstallerError{
+				ErrorCode: internal.OrchInstallerErrorCodeTerraform,
+				ErrorMsg:  fmt.Sprintf("cannot find password in %s module output", s.Name()),
+			}
+		}
```
yi-tseng-intel, line 296: Cam we use password_id instead of password?
The reason I was using password_id is that the master password will rotate every few days.

Given this scenario:
1. We install the orchestrator with RDS password version 1
2. Few days later, the password is rotated by RDS
3. In another release, we adds a new database. And stage-2 will try to create a new one during upgrade with RDS master password. in this case, we need to fetch latest password from secrets master with password_id.

Maybe we can include both password_id and password in the runtime state and we use password_id for aws and use password for on-prem?

new-installer/internal/steps/aws/rds_test.go
```diff
@@ -0,0 +1,263 @@
...
+			Output:         map[string]tfexec.OutputMeta{},
+		}, nil).Once()
+	}
+}
+
+func (s *RDSStepTest) expectAWSUtiliyCall(action string) {
+	if action == "install" {
+		s.awsUtility.On("GetAvailableZones", s.config.AWS.Region).
+			Return(availabilityZones, nil).
+			Times(2) // One for install, and one for uninstall
```
yi-tseng-intel, line 253: Once? Since you already have the `else if` below for uninstall
```suggestion
			Once()
```

new-installer/targets/aws/iac/rds/rds_test.go
```diff
@@ -0,0 +1,123 @@
...
+	s.NotEmpty(readerHost, "Expected RDS Reader Host to be created")
+
+	port := terraform.Output(s.T(), terraformOptions, "port")
+	s.NotEmpty(port, "Expected RDS Port to be created")
+
+	username := terraform.Output(s.T(), terraformOptions, "username")
+	s.NotEmpty(username, "Expected RDS Username to be created")
+
+	password := terraform.Output(s.T(), terraformOptions, "password")
+	s.NotEmpty(password, "Expected RDS Password to be created")
```
yi-tseng-intel, line 122: also check password_id

yi-tseng-intel (approved):

new-installer/internal/steps/aws/rds.go
charlesmcchan: This terraform output
```
output "password" {
  value     = jsondecode(data.aws_secretsmanager_secret_version.rds_master_password.secret_string)["password"]
  sensitive = true
}
```
always fetches the latest version of the secret. This ensures that the runtime state is updated every time Terraform is applied, including during upgrade scenarios:
```
  if password, ok := terraformStepOutput.Output["password"]; ok {
      runtimeState.Database.Password = strings.Trim(string(password.Value), "\"")
```
As long as Stage 1 is executed before Stage 2—where the runtime state is used to update the databases—this approach will work reliably.

The only potential exception is during daily operations. If something goes wrong and we need the DB master password for troubleshooting, but Stage 1 hasn’t been run, the password in the runtime state file may be outdated. However, this is acceptable, as the latest password can always be retrieved directly from AWS Secrets Manager in such edge scenario.

PR #804, opened 2025-09-23
added ~ chracter as prefix in os-profile revision to support dynamic addition of os-profile patched versions 

SushilLakra:
### Description

added ~ chracter as prefix in os-profile revision to support dynamic addition of os-profile patched versions for any os-profile during runtime without the orchestrator upgrade or redeployment

Fixes # (issue) ITEP-76300

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/.github/actions/cypress/action.yaml
+++ b/.github/actions/cypress/action.yaml
@@ -40,7 +40,7 @@ runs:
       with:
         repository: open-edge-platform/orch-ui
         path: orch-ui
-        ref: 108cfaaaa9ac5839d3fc9e57c3bc405ced02f9df
+        ref: d17f65821d4879b80300d942966ca919f2a51924
         persist-credentials: false
     - run: git log -1
       shell: bash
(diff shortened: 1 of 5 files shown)


argocd/applications/configs/infra-managers.yaml
```diff
@@ -69,7 +69,7 @@ telemetry-manager:
 
 os-resource-manager:
   managerArgs:
-    osProfileRevision: 0.8.2
+    osProfileRevision: ~0.8.2
```
Ram-srini, line 72: What does ~ meaning here?

SushilLakra: The ~ character means supporting range or multiple versions for an os-profile. These versions of os-profile will be downloaded automatically from Release Server as soon as it is available. Please note that these versions can only be different patch versions and it can't be any major/minor versions. 
For example, users are allowed to set the osProfileRevision in the configuration of the OS Resource manager, by either setting a precise version of the OSProfile or with a ~ prefix. For example, if the orchestrator is configured with 0.3.1, only OSProfile version 0.3.1 will be allowed in the orchestrator. If, instead, ~0.3.1 is set, all versions in the range will be picked >=0.3.1, <0.4.0.

Ram-srini (approved):

PR #816, opened 2025-09-24
App Orch Tenant Controller simplification ADR

scottmbaker:
### Description

This is an ADR for simplifying app-orch-tenant-controller.

### Checklist:

### DIFF
--- a/design-proposals/app-orch-tenant-controller-simplification.md
+++ b/design-proposals/app-orch-tenant-controller-simplification.md
@@ -65,7 +65,7 @@ The proposed format of the configmap is as follows:
 2. The configmap shall have the label `app-orch-tenant-controller-desired-state: present` if
    the deployment package should be loaded or `absent` if it should not be unloaded.
 
-3. The confimap shall contain a set of yaml files that contain the deployment-package, 
+3. The confimap shall contain a set of yaml files that contain the deployment-package,
    application, registry (if necessary), etc. These yaml files are identical to the yaml files
    that have been previously located in the cluster-extensions repo.
 


design-proposals/app-orch-tenant-controller-simplification.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Proposal 2: Specify packages using a Helm Chart in cluster extensions
+
+This is similar to Proposal 1, except that rather than incorporating the list of deployment packages
+directly into the app-orch-tenant-controller helm chart, a second helm chart is created in
+cluster-extensions that holds the configmap that contains the list of deployment packages.
+
+This approach allows a separation between code and data. The "code" remains in the app-orch-tenant-controller
+helm chart whereas the "data" is present in the cluster-extensions helm chart. Updating a new extension
+only requires generating the cluster-extensions helm chart, which contains only a set of yaml files.
```
hyunsun, line 62: As mentioned, this cluster-extensions helm chart would primarily contain YAML data files (deployment package specs), which deviates from typical Helm chart patterns. Could you provide more specifics on how will these YAML files be converted into K8s resources (ConfigMaps, Secrets, etc.)? Or how will the App Tenant Controller discover and consume these YAML files?

In case you need more options, here's one more. It's similar to the current approach but I think simpler and more flexible. We could make the App Tenant Controller configurable to source default deployment packages from either a URL or K8s resources:
```
defaultDeploymentPackages:
  type: "url" # or "configmap"
  url: "https://github.com/org/cluster-extensions/releases/latest/download/default-addons.tar.gz"
  # configMap:
  #  name: 
  # adding a flexibility to dynamically create a ConfigMap from helm value content might be perfect
     for users who want no additional steps to create a ConfigMap
```
Maybe it somewhat overlaps with your proposal 2 if you were thinking to make yamls into configmap. It just feels a bit unnatural for me to create a helm chart with data only, so I suggest using github releases to publish deployment packages as downloadables and make it our default.

PR #825, opened 2025-09-26
Fix the yamllint issue

adimoft:
### Description

Just trying to fix the yamllint issue.

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/.github/workflows/virtual-integration.yml
+++ b/.github/workflows/virtual-integration.yml
@@ -224,9 +224,6 @@ jobs:
       - name: Setup asdf and install dependencies
         uses: open-edge-platform/orch-utils/.github/actions/setup-asdf@main  # zizmor: ignore[unpinned-uses]
       
-      - name: Install yamllint properly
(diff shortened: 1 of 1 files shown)


.github/workflows/virtual-integration.yml
```diff
@@ -223,6 +223,9 @@ jobs:
 
       - name: Setup asdf and install dependencies
         uses: open-edge-platform/orch-utils/.github/actions/setup-asdf@main  # zizmor: ignore[unpinned-uses]
+      
+      - name: Install yamllint properly
+        run: pip install yamllint
```
pankajgoyal2, line 228: why is there a need to install yamllint seperately? Wasn't it getting installed via asdf?
Also, you are not specifying the version of yamllint to be used here?

adimoft: Yes, you're right. It should come from asdf.

PR #859, opened 2025-10-03
Split pre-installer from installer: On-prem

PalashGoelIntel:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


scottmbaker (dismissed):
Looks good to me. Let's make sure we also update documentation as necessary, for example https://docs.openedgeplatform.intel.com/edge-manage-docs/dev/deployment_guide/on_prem_deployment/on_prem_get_started/on_prem_install.html. Also, should try running onprem_installer.sh without running the pre-installer and seeing if the error messages are clear when the environment has not been setup.

PalashGoelIntel:
> Looks good to me. Let's make sure we also update documentation as necessary, for example https://docs.openedgeplatform.intel.com/edge-manage-docs/dev/deployment_guide/on_prem_deployment/on_prem_get_started/on_prem_install.html. Also, should try running onprem_installer.sh without running the pre-installer and seeing if the error messages are clear when the environment has not been setup.

i am checking is the kubernetes installed before running the installer script .
`if [ $(dpkg -l | grep -i onprem-ke-installer| wc -l) -eq 0 ]; then
    echo "Please run pre-installer script first"
    exit 1
fi
`

PR #897, opened 2025-10-14
Update edge node manifest version to latest

sys-orch:
Bumping edge node manifest version to latest iteration


rranjan3:
Taken care in #898

PR #1013, opened 2025-11-05
Fix for code scan issues

PalashGoelIntel:
https://github.com/open-edge-platform/edge-manageability-framework/security/code-scanning/22
https://github.com/open-edge-platform/edge-manageability-framework/security/code-scanning/23

### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/argocd/applications/configs/infra-managers.yaml
+++ b/argocd/applications/configs/infra-managers.yaml
@@ -75,7 +75,7 @@ telemetry-manager:
 
 os-resource-manager:
   managerArgs:
-    osProfileRevision: 0.8.9
+    osProfileRevision: 0.8.10
     osSecurityFeatureEnable: false
     inventoryAddress: "inventory.orch-infra.svc.cluster.local:50051"
     traceURL: "orchestrator-observability-opentelemetry-collector.orch-platform.svc:4318"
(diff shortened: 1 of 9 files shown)


mage/Magefile.go
```diff
@@ -1498,10 +1498,15 @@ func (l Lint) Yaml() error {
...
+// To retrieve the password, use:kubectl get secret orch-database-postgresql -n orch-database -o jsonpath='{.data.password}' | base64 --decode
 func (d Database) GetPassword() error {
-	pass, err := d.getPassword()
-	fmt.Println(pass)
-	return err
+	_, err := d.getPassword()
+	if err != nil {
+		return err
+	}
+	fmt.Println("kubectl get secret orch-database-postgresql -n orch-database -o jsonpath='{.data.password}' | base64 --decode")
```
scottmbaker: I don't see how providing the user the command to print the password is any more secure than just printing it ourselves. It may silence the code-scanning issue, but I don't feel like this is a valid issue. This command is the result of a direct user action (`mage database:GetPassword`). Asking the user to copy/paste a command is inefficient/inconvenient, when the end result is exactly the same as had we printed it ourselves.

PR #1131, opened 2025-11-24
ADR for Web UI Composability

vigneshintel:
### Description

Creating an ADR to explain how users can compose different micro-frontends in the UI. 

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


effndc:
I apprecaite the ADR discussing this implementation choice.  I was hesitant and had worried that this would require significant modification to the front end that could result in fragility, it sounds as though this just extends the current UI modularity and just avoid the forced pre-load.  

This seems like a good path to reduce our resource requirements for simpler deployments.

PR #1308, opened 2025-12-23
EIM API design

damiankopyto:
### Description

This PR provides detailed design for EIM decomposed API

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/design-proposals/eim-modular-api-workflows.md
+++ b/design-proposals/eim-modular-api-workflows.md
@@ -102,7 +102,7 @@ path/APIs/fields in inventory are used.
 The EIM APIv2 modularity will be achieved by enabling only necessary
 services based on required functionality by registering only necessary service handlers.
 Services are grouped
-together in combination to make up API sets tailored for a given workflow.
+together in manifests, forming API sets tailored for a given workflow.
 
 The services/APIs will be categorized into the following buckets:
 


design-proposals/eim-modular-api-workflows.md
```diff
@@ -0,0 +1,126 @@
...
+
+### EIM - vPRO only
+
+In this deployment only the minimum viable EIM API stack will be deployed along with the vPRO related APIs and resource managers and agents. In this scenario the EIM API will only expose the host services used to manage the host resource - since only the host resource is needed to onboard the EN, activate vPRO and do power management of the EN via vPRO.
+
+This version of modular EIM API will be deployed with the following service:
+
+- Onboarding service APIs
+
+It is assumed that provisioning/instance creation and creation of associated resources related to provisioning is not needed.
```
jkossak: It needs double checking if this scenario can work without instance creation. I can see that onboarding manager repeats operations on the instance resource. If the instance resource is missing, the onboarding manager may break unless its logic is adjusted to support onboarding without provisioning (i.e. no instance manipulation).

damiankopyto: Yes - the assumption made here was that we will have a "smaller" onboarding manager that will not do anything provisioning related, but let's check in with the team.

PR #1616, opened 2026-03-17
add ADR for vpro profile platform simplification

scottmbaker:
### Description

Adds a new ADR to describe options for simplifying the vpro profile's platform dependencies.

### Checklist:


design-proposals/platform-dns-simplification.md
```diff
@@ -0,0 +1,380 @@
...
+| `connect-gateway.example.com` | `example.com/connect/` | Straightforward (WebSocket upgrade on path) |
+| `fleet.example.com` | `example.com/fleet/` | Straightforward |
+
+**Edge-node-facing gRPC services** become path prefixes using gRPC's native path routing:
+
+| Current | Proposed |
+| --------- | ---------- |
+| `infra-node.example.com` | `example.com` with gRPC service routing |
+| `update-node.example.com` | `example.com` with gRPC service routing |
+| `telemetry-node.example.com` | `example.com` with gRPC service routing |
```
cjnolan, line 124: For telemetry, would there be separate service routes for logs and metrics, e.g. `example.com/logs` and `example.com/metrics`, or as a single `example.com/telemetry` route? Both go to the same OpenTelemetry Collector service in the Edge Node Observability pipeline in orch-infra and are received on the same [otlp receiver](https://github.com/open-edge-platform/o11y-charts/blob/main/charts/edgenode-observability/deployments/edgenode-observability/values.yaml#L83) so a single route would likely make the most sense and would further reduce the complexity.

design-proposals/platform-dns-simplification.md
```diff
@@ -0,0 +1,380 @@
...
+
+| Current | Proposed | Complexity |
+| --------- | ---------- | ------------ |
+| `web-ui.example.com` | `example.com/` | Straightforward |
+| `api.example.com` | `example.com/api/` | Already uses PathPrefix |
+| `keycloak.example.com` | `example.com/auth/` | Moderate (see below) |
+| `vault.example.com` | `example.com/vault/` | Straightforward |
+| `registry-oci.example.com` | `example.com/registry/` | Complex (see below) |
+| `observability-ui.example.com` | `example.com/grafana/` | Straightforward - Grafana supports root_url |
+| `observability-admin.example.com` | `example.com/grafana-admin/` | Straightforward |
```
cjnolan, line 107:
```suggestion
| `observability-ui.example.com` | `example.com/observability-edgenode/` | Straightforward - Grafana supports root_url |
| `observability-admin.example.com` | `example.com/observability-orchestrator/` | Straightforward |
```
Would these make more sense as it specifies what dashboards exactly each provides? Even with a switch to `/grafana` the difference between what the two services are covering is not clear

design-proposals/platform-dns-simplification.md
```diff
@@ -0,0 +1,380 @@
...
+
+Traefik already supports gRPC routing with `PathPrefix` rules matching gRPC
+service paths. Since each gRPC service has a unique protobuf package and service name,
+there are no path collisions.
+
+The change requires updating the `infra-config` ConfigMap
+(`infra-onboarding.tpl`) that tells edge nodes which hostnames to connect to.
+Instead of 10+ hostnames, edge nodes would receive a single hostname with
+different gRPC service paths. The edge node agents must be updated to connect to
+the new single hostname.
```
cjnolan, line 203: All of the agents get the service address past to them during install using debian settings provided by the installer script. These service paths are set by the `infra-config` configmap. Then during the agent package install, the postinst script in the package updates the service URL entries in the agent configuration file with the provided URL from the installer script (for example, see the [HW Discovery Agent postinst](https://github.com/open-edge-platform/edge-node-agents/blob/main/hardware-discovery-agent/debian/postinst#L8)).

If the `infra-config` configmap values are updated, these new service paths will be fed through directly to the agents during onboarding. Are there separate changes needed to the agents themselves for the gRPC configuration for the new service paths to work?

design-proposals/platform-dns-simplification.md
scottmbaker: I think these should remain as separate routes, with the logs route going to loki and the metrics route going to mimir.

design-proposals/platform-dns-simplification.md
scottmbaker: agreed.

design-proposals/platform-dns-simplification.md
scottmbaker: Adding a section to clarify. Will put in a separate PR.

design-proposals/platform-dns-simplification.md
scottmbaker: to be addressed in a separate PR.

design-proposals/platform-dns-simplification.md
scottmbaker: to be addressed in a separate PR.

johnoloughlin (approved):

cjnolan (approved):

PR #1624, opened 2026-03-18
Update vpro-profile-simplification.md

johnoloughlin:
ADR changes


scottmbaker (approved):
I'd like to adjust the language in the ingress-nginx statement slightly. Rather than "is deprecated", we can say "is not used in this profile".

PR #1665, opened 2026-03-31
ADR for modular observability pipeline

cjnolan:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,249 @@
...
+  [XPU System Mnagement Interface](https://github.com/intel/xpumanager) package to be installed on the edge node.
+- **Performance Monitoring Unit (PMU) Metrics**: These are metrics specific to Intel CPUs and can be read using the
+  [intel_pmu collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_pmu) in Telegraf. Currently,
+  the POA metrics service configures this plugin but does not enabled it by default.
+- **Cache Utilization and Performance Metrics**: The primary collector for this will be the [intel_rdt collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_rdt)
+  in Telegraf, which uses [Intel Resource Director Technology](https://github.com/intel/intel-cmt-cat) to report the
+  utilization of the L3 cache. As well as this collector, the intel_pmu collector above also provides some cache performance
+  metrics as does the [intel_pmt collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_pmt)
+  in Telegraf when used with newer Intel processors.
+- **BIOS Metrics**: This would require a new collector to provide such metrics from the BIOS environment.
```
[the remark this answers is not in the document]
cjnolan: Telegraf does have a redfish collector for temperature and power metrics. We can use both to report on the BIOS settings. Updated to outline this.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,249 @@
...
+  provide a variety of metrics. The [Telegraf disk collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/disk)
+  gathers utilization metrics while the [diskio collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/diskio)
+  reports the read and write counts to the edge node storage devices. There is also the
+  [smart collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/smart) which, when run on an
+  edge node that has storage devices that support it, will provide additional utilization metrics. Finally, the POA also
+  provides a [script](https://github.com/open-edge-platform/edge-node-agents/blob/main/platform-observability-agent/scripts/collect_disk_info.sh)
+  that can be run by the [exec plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/exec)
+  in Telegraf. Of the four collectors, the POA metrics service has configuration settings for all four, but only enables
+  the disk and diskio plugins by default.
+- **iGPU Utilization and Performance Metrics**: There is currently no plugin in Telegraf that gathers such metrics, so to
```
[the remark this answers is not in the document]
cjnolan: Updated to outline how these packages combined with the current metrics script for GPU metrics can be used to gather these metrics.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+For more details on the POA and the individual services it installs on the edge node, please see the
+[developer guide](https://docs.openedgeplatform.intel.com/edge-manage-docs/dev/developer_guide/observability/arch/orchestrator/edgenode-observability.html)
+for the agent.
+
+## Proposal
+
+To support collection of additional HW metrics from GPU, PMU, cache utilization, etc., the current POA implementation
+will be expanded to include new metrics collectors for these HW components. Also, modifications will be made to
+the Edge Node Observability pipeline deployment in the orchestrator to allow it to be deployable as a standalone
+pipeline without requiring other components from the EMF stack.
```
palade: >without requiring other components from the EMF stack

Which are these components?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+  which, when run on an edge node that has storage devices that support it, will provide additional utilization metrics.
+- **dGPU Utilization and Performance Metrics**: Currently, the POA metrics service provides a [script](https://github.com/open-edge-platform/edge-node-agents/blob/main/platform-observability-agent/scripts/collect_gpu_metrics.sh)
+  that can collect metrics from dGPU devices on an edge node. It requires the
+  [XPU System Management Interface](https://github.com/intel/xpumanager) package to be installed on the edge node.
+- **Performance Monitoring Unit (PMU) Metrics**: These are metrics specific to Intel CPUs and can be read using the
+  [intel_pmu collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_pmu) in Telegraf.
+- **BIOS Metrics**: One option for these metrics is to use the [Telegraf redfish collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/redfish)
+  to retrieve thermal and power settings.
+
+##### New Metrics to Configure and Enable
```
palade, line 159: Each of these will require additional permissions to be enabled on the device side and will likely increase resource utilization. Is there any QoS currently enabled to support existing and additional metrics collection? e.g.,  best-effort collection?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+for the service.
+
+#### Workflow Design
+
+For the pipeline, the pipeline will remain as it currently is when deploying the full EMF stack, however the
+modular workflow will not deploy the Grafana UI and dashboards when it is deployed without the full stack.
+Instead the Orchestrator Command Line Interface (CLI) tool will be extended to provide commands for a user
+to run to query the Mimir backend for metrics.
+
+The CLI will receive a command containing the metric to be queried for, the edge node to be checked as well as
```
palade: To clarify this, is the query going to the orchestrator or to the edge node? Or if the requested data is not found on the orchestrator side, then it will be sent to the edge node?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+any time range required by user. If a time range is not provided, then the CLI should use a default time range,
```
palade: Maybe a concern that is already addressed, but how is the clock synchronization is ensured across orchestrator and edge devices to ensure that the requested time range  is the same across devices and there is no offset?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+such as the last 5 minutes. The CLI should also support retrieving both averages and sums for metrics over set time
+periods.
```
palade: Is the request made for a single node or for all the nodes? how does this scale?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+
+Within the CLI, it should convert the received query into the PromQL format needed for querying Mimir
+and then send the PromQL query to the Mimir API. When the CLI receives the metrics back from Mimir,
```
palade: What happens if, for some reason, metrics collection fails or becomes unavailable during the requested time range? What if the data is only partially available?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+  - Identify the services needed for a modular observability workflow.
+  - Modify the deployment profiles to include an observability only modular workflow.
+  - Test the deployment of the new modular workflow.
+  - Test deployment with the updated edge node observability agent.
+  - Extend Orchestrator CLI to retrieve metrics from the observability pipeline.
+  - Test the updated CLI with the modular observability pipeline and confirm that new metrics can be retrieved.
+  - Provide documentation on how to install the modular observability workflow.
+  - Extend Orchestrator CLI documentation with new commands for metrics querying.
+
+## Opens
```
palade, line 524: Has support for multi-vendor environments been considered?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+    CLI ->> CLI: Convert metrics response from Mimir into user readable format
+    CLI ->> US: Return requested metrics in user readable format
+    deactivate CLI
+  end
+```
+
+## Implementation Plan
+
+- Hardware Metrics Collection.
+  - Identify the new hardware metrics collectors to be added to the current edge node metrics service.
```
palade, line 506: What are the performance implications when additional metrics are collected, both on the orchestrator side and across the network?

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+## Opens
+
+- Grafana dashboards will not be used in modular flow, instead metrics will be retreived using the CLI. Do
+  we still require the dashboards to be maintained?
+- Investigate the current support in CLI for retrieving metrics from Mimir
+- Not included in this proposal is the telemetry management pipeline that runs parallel to the observability
+  pipeline and can be used to configure what metrics an edge node reports after it has been deployed without
+  requiring a full redeployment or access to the edge node. For modular deployments, should this also be included
+  and used for this purpose or should it be exlcuded?
+- Investigate the [Intel Performance Counter Monitor(PCM)](https://github.com/intel/pcm) tool as there may be
```
palade, line 535: What happens if the hardware metrics collection fails? i.e., there is some hardware malfunction, or due to misconfigurations, or there is some disruption to sensors doing readings

palade (commented):
Left a few comments

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+
+#### Metric Collectors
+
+On the edge node, the POA metrics service currently provides a number of metrics by default as well as some that are
+configured but disabled by default. For this workflow, there are also additional metrics to be added.
+
+##### Configured and Enabled Metrics
+
+- **CPU Utilization and Performance Metrics**: CPU usage metrics can be retrieved using the [Telegraf cpu collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cpu).
+- **Memory Utilization and Performance Metrics**: The [Telegraf mem collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mem)
```
Ram-srini, line 130: Good capture memory utilization matrix for both virtual memory and UMA Unified memory/VRAM (used by iGPU) because AI workloads use both of these memory segments.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+- It will also cover how the pipeline and its deployment will be updated to work in a modular environment,
+  including any changes to be made to the current pipeline in order to support this.
+- Modularization changes outlined below are designed to work with all use cases outlined in the
+  [Modular Decomposition documentation](./eim-modular-decomposition.md).
+
+### Design
+
+#### Metric Collectors
+
+On the edge node, the POA metrics service currently provides a number of metrics by default as well as some that are
```
Ram-srini: Please check the possibility of having the k3s observability matrix enabled in FedAero EN which is stand-alone ubuntu with k3s cluster deployed.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+  Interface package needs to be installed on the edge node along with the intel-level-zero-gpu package. Using these packages
+  with the script currently used for dGPU metrics in the POA metrics service will allow it to also retrieve iGPU metrics.
+- **Cache Utilization and Performance Metrics**: The primary collector for this will be the [intel_rdt collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_rdt)
+  in Telegraf, which uses [Intel Resource Director Technology](https://github.com/intel/intel-cmt-cat) to report the
+  utilization of the L3 cache. As well as this collector, the intel_pmu collector above also provides some cache performance
+  metrics as does the [intel_pmt collector](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/intel_pmt)
+  in Telegraf when used with newer Intel processors.
+- **BIOS Metrics**: For other BIOS settings, dmidecode can be run using the Telegraf exec collector to gather these.
+- **VPU Utilization and Performance Metrics**: This will require a new collector to retrieve metrics from any
```
Ram-srini, line 175: what does VPU refer to? Is it a vision processing unit which is Image processing unit in intel?

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: The cluster and app orchestration and EIM, since the pipeline should be independent of these and should be deployable without needing them to be deployed. I'll update this

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: We can look into this but it should be possible to do so. The components used in the observability stack are the POA service on the edge node, which gathers the metrics, and the central observability stack that would be deployed in the orchestrator.

The agent can continue to be run on the edge node as a host service in this case and the pipeline deployed on the edge node cluster, with the endpoint address changed for the agent.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: The current memory collection is for the system memory usage, VRAM memory utilization would be collected as part of the iGPU metrics collection.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: They will need additional file permission access for the agent. There will be some resource utilization if enabling more metrics, however this is also currently an issue which is why we only enable certain metrics on start up. To modify these, we can use the platform telemetry agent and telemetry manager to set observability profiles for an edge node, which can control what metrics are collected. Since this is connected to the EIM, there is an open as to whether we want to include this as part of this effort.

Another option is allowing for modification of the metrics service configuration using orch-cli so that the metrics collection can be configured based on requirements. At install time, will enable a default selection of metrics for collection like we currently do. Will look into how this can made configurable

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: Yes but I don't think it is as necessary as including NPU, which was removed in error. Will modify to add NPU back and clarify importance.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: It will go to the metrics storage in Mimir, similar to how the current Grafana dashboards query for metrics when in use. So the query from CLI will go to the Mimir service, wherever it is deployed in the environment.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: I believe there is some NTP synchronization enabled between edge node and orchestrator already, but will check and confirm this.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: It should be for a single node, since the query will require the edge node name or ID. PromQL queries can be run across multiple nodes based on the labels searched for but that can make the output hard to understand.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+  VPUs on an edge node.
+
+To view the current POA metrics service configuration, please see the [configuration file](https://github.com/open-edge-platform/edge-node-agents/blob/main/platform-observability-agent/configs/poa-telegraf.conf)
+for the service.
+
+#### Workflow Design
+
+For the pipeline, the pipeline will remain as it currently is when deploying the full EMF stack, however the
```
Ram-srini: Please call out the deployment support will be for onprem or cloud.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: If they are unavailable during the selected time period, the result from Mimir would be an empty metrics series. For partial metrics, then it will return those as the result and for average and sum queries, it will use those partial metrics.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,256 @@
...
+
+- Hardware Metrics Collection.
+  - Identify the new hardware metrics collectors to be added to the current edge node metrics service.
+  - Extend the current GPU metrics script to also collect iGPU metrics using the Telegraf exec plugin.
+  - Develop a new collector to retrieve VPU metrics.
+  - Add additional Telegraf plugins to metrics service configuration.
+  - Test deployment of updated metrics sevice on edge node and check the metrics being retrieved.
+  - Update documentation for the edge node observability agent.
+- Modular observability workflow.
+  - Identify the services needed for a modular observability workflow.
```
Ram-srini: For the on-prem deployment of observability components specify the system resource requirements and list of port to be opened if it is node port.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: For the network, we send using batching so batch size will not change but there may be more messages sent. On orchestrator, there would be some more processing and more metrics being stored. The amount would depend on what metrics are enabled and the settings used for retrieving them using the metrics service (frequency, full metric set or subset, etc.).

It is the same as the current considerations we have with observability, not all metrics are enabled on start up in POA, although more can be enabled if needed. However, there will be an increase in storage needed and processing if enabled. It is on the user and the use case being tested that will determine if these additional metrics are collected and if the performance impact is acceptable.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: As in for different OS environments? No but we are using open source metrics collectors and services for the implementation and these have been tested in different environments.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: Then the metrics service will log an error and will not provide metrics for that plugin only. However, it will not impact other metrics from being collected that are not impacted by the error, unless the service itself errors and stops.

design-proposals/observability-modular-hw-metrics-collection.md
jkossak: the portion of UMA currently allocated to iGPU will be captured as part of the iGPU metrics. VRAM utilization would be rather collected for dGPU.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: Will add this

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: System resources should remain the same as currently used but will add here. Will also add port information.

design-proposals/observability-modular-hw-metrics-collection.md
jkossak: I guess for modular deployment we can only assume the clock synchronization between orchestrator and edge nodes.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,536 @@
...
+### ObaaS Pipeline
+
+In the orchestrator, the ObaaS pipeline also uses open source components to process and store
+metrics from the connected edge nodes. The services it runs includes:
+
+1. [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) which has been configured
+   to apply and configure labels and metadata for filtering edge node metrics as well as for supporting
+   multitenancy environments.
+2. [Grafana Loki](https://github.com/grafana/loki) is used for the logs backend and storage.
+3. [Grafana Mimir](https://github.com/grafana/mimir) is used for the metrics backend and storage.
```
scottmbaker, line 67: Is it possible to opt for a more lightweight solution than Mimir? For example, can we run Prometheus instead? and if we did so, would it reduce the footprint on the orchestrator? Perhaps as future work if not doable for this release.

design-proposals/observability-modular-hw-metrics-collection.md
```diff
@@ -0,0 +1,536 @@
...
+queried as well as an query specific flag to be included.
+
+Within the CLI, it should convert the received query into the PromQL format needed for querying Mimir
+and then send the PromQL query to the Mimir API. When the CLI receives the metrics back from Mimir,
+it should convert it into a easy read format before returning it to the user. If there are no metrics
+found in Mimir, then an empty result will be returned but if there are partial results over a range
+it will provide those partial results. For more details on how to create and use PromQL queries, see
+the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/basics/).
+
+The CLI commands will look like the following:
```
scottmbaker, line 283: Please check with @damiankopyto  regarding his current CLI refactoring. He is working on adding flexible table formatting, yaml, and json output. It's possible we may be able to apply his work to the data you're reporting.

scottmbaker (approved):

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: It would need to be investigated but I think Prometheus would give similar support for storage and querying metrics that we get from Mimir. Not sure how much of an impact on the orchestrator footprint it would have, so that would need to be checked. It is something that can be looked at in the future.

design-proposals/observability-modular-hw-metrics-collection.md
cjnolan: I'll sync with him on this. I was aware of it from the ADR he has created and, since the output from Mimir when running a query is in a JSON format, it should feed in well to that.

PR #1689, opened 2026-04-09
add adr on dns simplification

scottmbaker:
### Description

ADR to simplify the number of names that are required for an orchestrator.

### Checklist:

### DIFF
--- a/design-proposals/platform-dns-simplification.md
+++ b/design-proposals/platform-dns-simplification.md
@@ -1,0 +1,380 @@
+# Design Proposal: DNS Simplification
+
+Author(s): Scott Baker
+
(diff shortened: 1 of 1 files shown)


johnoloughlin (approved):

palade (approved):

sunil-parida:
Yes @scottmbaker adr LGFM using (*.CLUSTER_FQDN) instead of adding individual service FQDNs for  single-node mode, only tinkerbell-haproxy-<CLUSTER_FQDN> should be mapped to the HAProxy load balancer, and *.CLUSTER_FQDN should be mapped to the Traefik load balancer for multimode.

sunil-parida (approved):

